Issue #129, opened 2026-06-28
Auto-download while playing + offline playback


jaxkodex:
> **Prerequisite:** do after #130 (PlayerController seam refactor), which establishes the queue/stream plug points this feature uses. Can then run in parallel with the sibling follow-up.

Implement the **auto-download** feature teased throughout the UI ("Saves as you listen", "Offline ready", the Downloads screen, and the Now Playing "SAVING OFFLINE" label).

## Goal
As tracks play, save the original files locally so the library becomes available offline ("local-first").

## Scope
- Download the **original** file bytes while streaming (the foundation already uses the `static` audio stream — `getAudioStreamUrl(static=true)` — so the streamed bytes are the file we want to keep).
- Persist downloaded-track metadata + file paths in a local store (Room is already in the version catalog).
- Serve playback from the local file when present; fall back to streaming otherwise.
- Wire the real Downloads screen (currently `SampleData`), the Home "On your device" section, the "Offline ready" pill, and the Now Playing "SAVING OFFLINE · %" indicator.
- Settings toggle to enable/disable auto-download and manage storage.
- Foreground/background download handling and storage-quota awareness.

## Foundation already in place (PR #127)
- `JellyfinRepository.audioStreamUrl` (static original-file stream)
- `PlayerController` / `PlaybackService`
- Manual DI container to host a download manager + Room DB

## Out of scope
- Selective/manual per-album downloads (this issue is the automatic-on-play path)

Issue #128, opened 2026-06-28
Random Walk: auto-generated continuous shuffle playlist


jaxkodex:
> **Prerequisite:** do after #130 (PlayerController seam refactor), which establishes the queue/stream plug points this feature uses. Can then run in parallel with the sibling follow-up.

Implement the **Random Walk** feature teased on the Home screen ("Press play, get on with life — a shuffled walk through your whole library") and in the Now Playing "PLAYING FROM" label.

## Goal
A one-tap, never-ending shuffled queue across the user's entire Jellyfin music library.

## Scope
- Generate an initial shuffled batch of audio items (e.g. `getItems(includeItemTypes=[AUDIO], sortBy=[RANDOM], recursive=true, limit=N)`).
- Continuously append more random tracks as playback nears the end of the queue (lookahead refill) so it never runs out.
- Wire the Home `RandomWalkCard` play button and the Now Playing "PLAYING FROM → Random Walk" source label to this mode.
- Avoid immediate repeats within a reasonable window.

## Foundation already in place (PR #127)
- `JellyfinRepository` (add a `randomTracks(limit)` query)
- `PlayerController.play(queue, startIndex)` + queue management
- Now Playing screen + ViewModel

## Out of scope
- Smart/seeded recommendations (pure random is fine for v1)

PR #131, opened 2026-06-28
refactor(playback): add stream-resolver and queue-refill seams (#130)

### LINKED ISSUE #130: Refactor PlayerController to add stream-resolver + queue-refill seams (unblocks #128/#129)
Small prep refactor so the two follow-ups (#128 Random Walk, #129 Auto-download) can be built in parallel without colliding in `PlayerController`.

## Why
`PlayerController.play()` and its queue logic are the only real overlap between #128 and #129. Introducing two indirection points now lets each feature plug into its own seam instead of both editing `play()`.

## Scope
In `data/playback/PlayerController.kt`:

1. **Stream-URL resolver indirection** — replace the direct `repository.audioStreamUrl(id)` call in `play()` with an injectable resolver, e.g.
   ```kotlin
   var streamUrlResolver: (Track) -> String? = { it.id?.let(repository::audioStreamUrl) }
   ```
   Default keeps today's behavior (remote static stream). #129 swaps in "local file if downloaded, else stream" without touching `play()`.

2. **Queue-refill hook** — add a callback the controller invokes when playback nears the end of the queue (e.g. on `onMediaItemTransition` / when remaining items < threshold):
   ```kotlin
   var onQueueRunningLow: (suspend () -> List<Track>)? = null
   ```
   Controller appends the returned tracks via `addMediaItems`. Default null = no refill (current behavior). #128 plugs its random-append logic here.

3. Keep `tracksById` population working for any items appended through the refill hook.

## Acceptance
- No behavior change with defaults (existing playback still works end to end).
- `assembleDebug` + `detekt` pass.
- Both seams are unit-coverable (resolver swap, refill append).

## Notes
- Branch off `feat/jellyfin-foundation` (PR #127) — or `main` once #127 merges.
- After this lands, #128 and #129 can proceed in parallel.

jaxkodex:
Closes #130.

Small prep refactor so #128 (Random Walk) and #129 (Auto-download) can be built in parallel without colliding in `PlayerController.play()`.

## Changes (`data/playback/PlayerController.kt`)

1. **Stream-URL resolver indirection** — `play()` no longer calls `repository.audioStreamUrl(id)` directly. It goes through an injectable property:
   ```kotlin
   var streamUrlResolver: (Track) -> String? = { it.id?.let(repository::audioStreamUrl) }
   ```
   Default keeps today's remote static-stream behavior. #129 swaps in "local file if downloaded, else stream" without touching `play()`.

2. **Queue-refill hook** — invoked from `onMediaItemTransition` when fewer than `REFILL_THRESHOLD` (3) items remain:
   ```kotlin
   var onQueueRunningLow: (suspend () -> List<Track>)? = null
   ```
   The controller appends the returned tracks via `addMediaItems`. Default `null` = no refill (current behavior). A `refilling` guard prevents re-entrant appends. #128 plugs its random-append logic in here.

3. **`tracksById` population** — `play()` and the refill path share a `buildItems()` helper that registers every resolved track in `tracksById`, so `nowPlaying` resolves correctly for items appended via the refill hook.

## Testability

Extracted two pure, Android-free helpers so both seams are unit-coverable:
- `resolvePlayables(tracks, resolver)` — resolver seam
- `shouldRefillQueue(currentIndex, itemCount, threshold)` — refill trigger

Added `PlayerControllerSeamsTest` covering resolver swap (custom URLs, dropping idless/unresolvable tracks) and the refill threshold.

## Verification

- No behavior change with defaults — existing playback path unchanged.
- `./gradlew testDebugUnitTest detekt assembleDebug` all pass.

### DIFF
--- a/app/src/main/java/pe/net/libre/mixtapehaven/data/playback/PlayerController.kt
+++ b/app/src/main/java/pe/net/libre/mixtapehaven/data/playback/PlayerController.kt
@@ -32,8 +32,21 @@ class PlayerController(
     private val appContext = context.applicationContext
     private val scope = CoroutineScope(Dispatchers.Main.immediate + SupervisorJob())
 
+    /**
+     * Resolves the playable URL for a [Track]. Defaults to the remote static stream from
+     * [repository]; swap it to source local files (e.g. downloads) without touching [play].
+     */
+    var streamUrlResolver: (Track) -> String? = { it.id?.let(repository::audioStreamUrl) }
+
+    /**
+     * Invoked when playback nears the end of the queue (fewer than [REFILL_THRESHOLD] items
+     * remain). The returned tracks are appended to the queue. Null disables refilling.
+     */
+    var onQueueRunningLow: (suspend () -> List<Track>)? = null
+
     private var controller: MediaController? = null
     private var pendingAction: (() -> Unit)? = null
+    private var refilling = false
     private val tracksById = mutableMapOf<String, Track>()
 
     private val _nowPlaying = MutableStateFlow<Track?>(null)
@@ -55,6 +68,7 @@ class PlayerController(
 
         override fun onMediaItemTransition(mediaItem: MediaItem?, reason: Int) {
             updateNowPlaying(mediaItem)
+            maybeRefillQueue()
         }
 
         override fun onPositionDiscontinuity(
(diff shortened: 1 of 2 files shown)


jaxkodex:
Thanks for the review — addressed all four points in a0e2b0a:

- **refillJob + cancel on play()**: the refill coroutine is now stored in `refillJob` and cancelled at the start of `play()`, so an in-flight refill can't append previous-queue tracks to a freshly started queue.
- **Crash-safe refill**: the hook is wrapped in `runCatching` (matching the repo's existing idiom) — failures are logged via `Log.w` instead of surfacing on the main thread, and `CancellationException` is rethrown to preserve coroutine cancellation.
- **Empty-queue guard**: `shouldRefillQueue` now returns `false` when `itemCount == 0`, so a stopped/cleared player doesn't immediately refill itself.
- **Test**: added a `shouldRefillQueue never refills an empty queue` case.

`testDebugUnitTest`, `detekt`, and `assembleDebug` all pass.

PR #139, opened 2026-07-04
Add devcontainer config for NAS dev environment

jaxkodex:
Adds `.devcontainer/` so the repo can run in the NAS dev-container launcher (and any devcontainer-compatible tool).

- JDK 17 base image (AGP 9.2.1 / Gradle 9.6.1 compatible)
- `post-create.sh` installs Android cmdline-tools, platform 36, build-tools 36 into a named volume (idempotent across rebuilds)
- Gradle cache on a named volume

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,25 @@
+{
+  "name": "mixtapehaven-android",
+  "image": "mcr.microsoft.com/devcontainers/java:17-bookworm",
+  "containerEnv": {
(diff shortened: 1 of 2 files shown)


jaxkodex:
Addressed all three review comments in 6ce400d:
- Now installs `platforms;android-37.0` + `build-tools;37.0.0` to match `compileSdk` 37 (note: platform 37 ships under the minor-versioned name `android-37.0`, not `android-37`)
- `mkdir -p $ANDROID_HOME` before the `chown`
- Stale `cmdline-tools/latest` is removed before the `mv` to prevent nesting

PR #148, opened 2026-07-17
feat: stream movies & TV series from Jellyfin (video step 1)

jaxkodex:
## Summary

First slice of video support: browse and **stream** movies & TV series from the Jellyfin server. Streaming-only by design — the manual mark-for-download pipeline (quality-capped transcodes) comes in a later step and plugs into a seam this PR leaves in place.

- **Data**: `VideoItem` model; repository queries for movies/series listing, item detail (with the user's resume position), and series episodes; `videoStreamCandidates()` returns ordered sources — direct play first, HLS h264/aac transcode fallback; `reportVideoPlayback()` sends started/progress/stopped playstate so the server resume point (and Continue Watching) stays in sync. DTO mapping lives in `JellyfinVideoMapping.kt`.
- **Home**: new "Movies & shows" horizontal poster rail (`PosterCard`, 2:3 portrait). Hidden when the server has no video items; a video fetch failure degrades to a hidden section and can't break the music Home.
- **Video Detail** (`video_detail/{itemId}`): backdrop with back overlay, title/meta, Play/Resume, overview, and an episode list with resume progress bars for series. For a series, Play targets the first in-progress episode.
- **Video Player** (`video_player/{itemId}`): second ExoPlayer instance owned by the ViewModel (survives rotation), media3 `PlayerView` transport controls, pauses music on start via new `PlayerController.pause()` (queue preserved). Resumes from the server position; on playback error falls through to the next stream candidate at the same position. Sources resolve via `AppContainer.videoSourceResolver` — the future hook for serving local downloads first, mirroring the audio `streamUrlResolver`.
- **Deps**: `media3-ui`, `media3-exoplayer-hls`.

## Test plan

- [x] `gradlew detekt` and unit tests pass
- [x] Verified on device (Pixel) against `https://demo.jellyfin.org/stable`: Home shows the poster rail with real artwork/meta → Dracula detail shows backdrop/overview and **Resume** → playback direct-plays, resumed from the server-side position (12:32 / 1:14:26), transport controls work → back reports stopped; re-entering resumes from a server position again
- [x] No video libraries / fetch failure: section hidden, music Home unaffected
- [ ] HLS fallback path (needs a file the device can't direct-play)

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -70,6 +70,8 @@ dependencies {
     implementation(libs.slf4j.api)
     implementation(libs.okhttp)
     implementation(libs.media3.exoplayer)
+    implementation(libs.media3.exoplayer.hls)
+    implementation(libs.media3.ui)
     implementation(libs.media3.session)
     implementation(libs.media3.datasource.okhttp)
     implementation(libs.coil.compose)
--- a/app/src/main/java/pe/net/libre/mixtapehaven/data/jellyfin/JellyfinRepository.kt
+++ b/app/src/main/java/pe/net/libre/mixtapehaven/data/jellyfin/JellyfinRepository.kt
@@ -7,20 +7,31 @@ import org.jellyfin.sdk.Jellyfin
 import org.jellyfin.sdk.api.client.ApiClient
 import org.jellyfin.sdk.api.client.extensions.audioApi
 import org.jellyfin.sdk.api.client.extensions.authenticateUserByName
+import org.jellyfin.sdk.api.client.extensions.dynamicHlsApi
 import org.jellyfin.sdk.api.client.extensions.imageApi
 import org.jellyfin.sdk.api.client.extensions.itemsApi
+import org.jellyfin.sdk.api.client.extensions.playStateApi
+import org.jellyfin.sdk.api.client.extensions.tvShowsApi
 import org.jellyfin.sdk.api.client.extensions.userApi
+import org.jellyfin.sdk.api.client.extensions.userLibraryApi
+import org.jellyfin.sdk.api.client.extensions.videosApi
 import org.jellyfin.sdk.model.api.BaseItemDto
 import org.jellyfin.sdk.model.api.BaseItemKind
 import org.jellyfin.sdk.model.api.ImageType
+import org.jellyfin.sdk.model.api.ItemFields
 import org.jellyfin.sdk.model.api.ItemSortBy
+import org.jellyfin.sdk.model.api.PlayMethod
 import org.jellyfin.sdk.model.api.SortOrder
 import org.jellyfin.sdk.model.api.request.GetItemsRequest
 import pe.net.libre.mixtapehaven.data.session.Session
 import pe.net.libre.mixtapehaven.data.session.SessionStore
 import pe.net.libre.mixtapehaven.model.Album
 import pe.net.libre.mixtapehaven.model.Track
+import pe.net.libre.mixtapehaven.model.VideoItem
 import java.util.UUID
+
+/** Lifecycle moment of video playback reported to the server via [JellyfinRepository.reportVideoPlayback]. */
+enum class VideoPlaybackEvent { STARTED, PROGRESS, STOPPED }
 
 /** Wraps the Jellyfin Kotlin SDK: authentication, library browsing, search, and stream/image URLs. */
 class JellyfinRepository(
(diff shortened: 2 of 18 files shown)


jaxkodex:
# Code Review

## Overview

Adds streaming-only video support: a "Movies & shows" poster rail on Home, a detail screen (backdrop, Play/Resume, episode list with progress bars), and a full-screen player backed by a second ExoPlayer owned by its ViewModel. The repository gains video queries, ordered stream candidates (direct play → HLS h264/aac fallback), and playstate reporting. A `videoSourceResolver` seam in `AppContainer` anticipates the download step. Well-scoped slice with a clear description and honest test plan.

## What's good

- **Architecture is consistent with the codebase**: repository → ViewModel → Compose screen, `appViewModel` factory, `runCatching` degradation matching the existing Home patterns. The `videoSourceResolver` seam mirrors the audio `streamUrlResolver` deliberately.
- **Failure isolation on Home** is done right: a video fetch failure degrades to a hidden section and can't break the music path (`HomeViewModel.kt`).
- **Candidate fallback in the player** (`VideoPlayerViewModel.onPlayerError`) keeps the reached position — a nice touch.
- The `HomeScreen` refactor into `HomeSections`/`HomeOverlays` is behavior-preserving and improves readability.
- Documentation comments explain *why* (e.g. the unhyphenated `mediaSourceId` convention, the `onCleared` scope note).

## Issues

### Correctness / behavior

- **No audio focus handling on the video player** (`VideoPlayerViewModel.kt`). The raw `ExoPlayer.Builder(...).build()` never requests audio focus. `musicController.pause()` handles this app's own music, but other apps (Spotify, podcasts) will keep playing over the video, and the video won't pause for calls/notifications. Add `setAudioAttributes(AudioAttributes..., /* handleAudioFocus = */ true)` on the builder — one line, and it gets both behaviors for free.
- **Playback continues when the app is backgrounded**. Nothing pauses the player on lifecycle stop — the screen has no `LifecycleEventObserver` and the ViewModel outlives visibility. With no foreground service or notification, video audio keeps playing invisibly after Home-button, and the 10s progress-report loop keeps the radio awake — the same class of idle battery drain fixed in #145. At minimum pause on `ON_STOP`; also consider stopping the report loop while paused (a `paused=true` ping every 10s indefinitely is unnecessary network churn).
- **`PlayMethod.DIRECT_STREAM` is reported unconditionally** (`JellyfinRepository.reportVideoPlayback`). When the player has fallen through to the HLS candidate the truthful value is `TRANSCODE`, so server-side stats will be wrong. The repository doesn't know which candidate won — either pass the play method from the ViewModel or note it as a known limitation.
- **STOPPED report can carry a stale position** (`VideoPlayerViewModel.onCleared`). If the source never became ready (both candidates failed), `currentPosition` is 0 and STOPPED is reported at 0 for an item the user may have half-watched — the STARTED→PROGRESS report at `resumeMs` mitigates this, but a failed session could still regress the resume point. Consider skipping the STOPPED report when playback never reached `STATE_READY`.
- **Series "next episode" logic** (`VideoDetailViewModel.playTarget`): `firstOrNull { resumePositionMs > 0 }` finds an *in-progress* episode, but once an episode is watched to completion Jellyfin zeroes its position, so a user who finished E1 gets sent back to E1 instead of E2. `tvShowsApi.getNextUp` solves exactly this; fine to defer, but worth a TODO.

### Minor

- `HomeViewModel.load()` now fetches videos and albums **sequentially**, adding video latency to the album load. Wrap them in `async` so the music Home isn't slowed by the new section.
- `VideoPlayerScreen`'s `AndroidView` never detaches `playerView.player` (`onRelease = { it.player = null }`) — a `PlayerView` can otherwise hold the surface briefly across disposal.
- The STARTED report fires even when the URL turns out unplayable — acceptable, just noting it's why the failed-session STOPPED case above matters.
- `EpisodeThumb` uses `posterUrl` (PRIMARY image), which for episodes is the 16:9 still — correct, but the 2:3-named field holding a 16:9 image is slightly confusing; a comment on `VideoItem.posterUrl` would help.

### Security

- `ApiKey` as a query parameter on stream/image URLs is the established pattern here, but the token now also rides on video URLs that ExoPlayer may include in playback error messages surfaced to `_error.value`. Low risk, no action needed now; a header-based `OkHttpDataSource.Factory` would remove it later.

### Test coverage

- No new tests. `formatRuntime`, `videoColorFor`, `toVideoItem` mapping, and especially `playTarget()` (the series-resume selection) are pure and cheap to unit test — `playTarget` is where the next behavioral bug will land, so I'd add those before merging.

## Verdict

Solid, well-structured first slice — the seams for the download step are thoughtfully placed. I'd fix **audio focus** and **background playback/polling** before merge (the second directly undoes the intent of the #145 battery fix); the rest can be fast-follows or TODOs.

jaxkodex:
Addressed the review in 4044edb:

**Must-fix**
- **Audio focus**: the video ExoPlayer now sets `AudioAttributes` (USAGE_MEDIA / CONTENT_TYPE_MOVIE) with `handleAudioFocus = true` — other apps' audio pauses/ducks, and calls pause the video.
- **Background playback**: `VideoPlayerScreen` observes lifecycle and pauses on `ON_STOP` (no background video service exists), and the progress loop now sends exactly one `paused=true` report per pause then stays quiet — no more indefinite 10s pings while paused, preserving the #145 battery intent.

**Also fixed**
- Playstate reports `TRANSCODE` when the HLS fallback candidate is active (`transcoding` flag threaded through `reportVideoPlayback`).
- STOPPED is skipped for sessions that never reached `STATE_READY`, so a failed session can't regress a resume point to 0.
- `AndroidView` `onRelease` detaches `playerView.player`.
- Home fetches videos and albums in parallel (`async`) — the new section adds no latency to the album load.
- `selectPlayTarget()` extracted as a pure function with unit tests (`SelectPlayTargetTest`), plus `formatRuntime`/`videoColorFor` coverage; the finished-episode → `getNextUp` limitation is documented on the function.
- `VideoItem.posterUrl` doc now notes it's a 16:9 still for episodes.

**Deferred (as suggested)**: `getNextUp` for series resume, header-based auth via `OkHttpDataSource.Factory`.

Local `detekt`, unit tests, and `lintDebug` all pass. Device re-verification of the ON_STOP/audio-focus behavior pending (phone currently off adb).

jaxkodex:
Device verification of the review fixes (Pixel, demo server):

- **ON_STOP pause**: with Dracula playing, pressing Home logged `player piid:76071 event:paused` in `dumpsys audio` within a second — no invisible background audio, no idle progress churn.
- **Audio focus**: the focus stack shows our entry held via media3's `AudioFocusManager` — `pack: pe.net.libre.mixtapehaven -- gain: GAIN -- attr: usage=USAGE_MEDIA content=CONTENT_TYPE_MOVIE`.

PR #149, opened 2026-07-18
feat: manual video downloads with quality-capped transcodes (video step 2)

jaxkodex:
## Summary

Movies and episodes can now be saved for offline playback. Downloads are **user-triggered** (no auto-download for video, unlike audio) and fetched as h264/aac **MPEG-TS transcodes** capped at the quality chosen in Settings (1080p / 720p / 480p, default 720p), so a phone-sized copy is stored instead of the original file.

- **Video Detail**: a Download button for movies/episodes and per-episode download icons for series, cycling Download → Downloading (streamed-bytes label, tap to cancel) → Downloaded (tap to remove).
- **Offline playback for free**: `AppContainer.videoSourceResolver` (the seam from step 1, #148) serves the saved local file as the only candidate when one exists, else falls back to the remote direct-play/HLS candidates. Zero player changes.
- **Downloads screen**: new "Movies & shows" section (poster, series/quality/size meta, per-item cancel/delete, tap-to-play), two-segment storage bar (music + video), and Remove all now clears both libraries.
- **Data layer**: `DownloadedVideo` entity + DAO, DB v1→v2 **additive migration** (existing audio downloads survive), `VideoDownloadManager` mirroring the audio manager (partial-file cleanup, 500 MB free-space guard, in-flight rows).

## Bugs caught by on-device verification (fixed in the second commit)

1. **Progressive mp4 transcodes are unplayable once saved**: ffmpeg cannot backpatch the `mdat` size over a non-seekable HTTP response, so the `moov` index at the tail is unreachable and ExoPlayer rejects the file as malformed. Switched the download container to MPEG-TS, which needs no finalization.
2. **Downloads died on screen-off**: Android freezes the cached process (`SocketException: Software caused connection abort`). A minimal dataSync foreground service is now held while downloads are active.
3. **Downloaded videos were unreachable offline** (Home rail and Detail are server-backed): Downloads-screen video rows are now tappable to play.

## Verification (Pixel 8, own server)

- Set quality to 480p via the new Settings dialog; verified persistence.
- Downloaded Before Sunset (1h 20m → 1.0 GB TS) **with the screen off the entire ~4 min** — dataSync FGS confirmed via `dumpsys activity services` (`isForeground=true`).
- Airplane mode → Downloads → tap row → movie plays from the local file (screenshot + fresh `event:started` in `dumpsys audio`).
- Cancel/remove paths exercised (removed the earlier broken mp4 via the Downloaded · Remove button; partial cleanup confirmed on abort).
- `test`, `detekt`, `lintDebug` all green; DB migration verified against a live v1 install.

## Deferred to step 3

- Offline Home view for downloaded movies (rail backed by the local library; offline taps route straight to the player).
- `tvShowsApi.getNextUp` for series resume; header-based stream auth instead of `?ApiKey=`.

### DIFF
--- a/app/src/main/java/pe/net/libre/mixtapehaven/data/download/DownloadDatabase.kt
+++ b/app/src/main/java/pe/net/libre/mixtapehaven/data/download/DownloadDatabase.kt
@@ -4,20 +4,38 @@ import android.content.Context
 import androidx.room.Database
 import androidx.room.Room
 import androidx.room.RoomDatabase
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
 
-/** Room database holding the offline download library. */
-@Database(entities = [DownloadedTrack::class], version = 1, exportSchema = false)
+/** Room database holding the offline download library (audio tracks and videos). */
+@Database(entities = [DownloadedTrack::class, DownloadedVideo::class], version = 2, exportSchema = false)
 abstract class DownloadDatabase : RoomDatabase() {
 
     abstract fun downloadDao(): DownloadDao
 
+    abstract fun videoDownloadDao(): VideoDownloadDao
+
     companion object {
+        /** v1 -> v2: adds the video download table. Additive so existing audio downloads survive. */
+        private val MIGRATION_1_2 = object : Migration(1, 2) {
+            override fun migrate(db: SupportSQLiteDatabase) {
+                db.execSQL(
+                    "CREATE TABLE IF NOT EXISTS `downloaded_videos` (" +
+                        "`id` TEXT NOT NULL, `title` TEXT NOT NULL, `kind` TEXT NOT NULL, " +
+                        "`seriesName` TEXT, `seasonEpisodeLabel` TEXT, `runtimeLabel` TEXT NOT NULL, " +
+                        "`posterUrl` TEXT, `artColorArgb` INTEGER NOT NULL, `qualityLabel` TEXT NOT NULL, " +
+                        "`filePath` TEXT NOT NULL, `sizeBytes` INTEGER NOT NULL, `complete` INTEGER NOT NULL, " +
+                        "PRIMARY KEY(`id`))",
+                )
+            }
+        }
+
         /** Build the app's download database in the default location. */
         fun build(context: Context): DownloadDatabase =
             Room.databaseBuilder(
                 context.applicationContext,
                 DownloadDatabase::class.java,
                 "downloads.db",
-            ).build()
+            ).addMigrations(MIGRATION_1_2).build()
     }
 }
--- a/app/src/main/java/pe/net/libre/mixtapehaven/data/download/DownloadDatabase.kt
+++ b/app/src/main/java/pe/net/libre/mixtapehaven/data/download/DownloadDatabase.kt
@@ -8,7 +8,7 @@ import androidx.room.migration.Migration
 import androidx.sqlite.db.SupportSQLiteDatabase
 
 /** Room database holding the offline download library (audio tracks and videos). */
-@Database(entities = [DownloadedTrack::class, DownloadedVideo::class], version = 2, exportSchema = false)
+@Database(entities = [DownloadedTrack::class, DownloadedVideo::class], version = 2, exportSchema = true)
 abstract class DownloadDatabase : RoomDatabase() {
 
     abstract fun downloadDao(): DownloadDao
--- a/app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadManager.kt
+++ b/app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadManager.kt
@@ -50,7 +50,8 @@ class VideoDownloadManager(
     private val scope: CoroutineScope = CoroutineScope(Dispatchers.IO + SupervisorJob()),
     private val client: OkHttpClient = OkHttpClient(),
 ) {
-    private val downloadsDir: File = File(context.applicationContext.filesDir, DIR).apply { mkdirs() }
+    private val appContext = context.applicationContext
+    private val downloadsDir: File = File(appContext.filesDir, DIR).apply { mkdirs() }
 
     /** Saved + in-flight video downloads, observed by the Downloads/Detail UI. */
     val downloads: Flow<List<DownloadedVideo>> = dao.observeAll()
@@ -96,6 +97,9 @@ class VideoDownloadManager(
         if (completedPaths.containsKey(id)) return
         synchronized(activeJobs) {
             if (activeJobs.containsKey(id)) return
+            // Foreground service for the duration: Android freezes cached processes on screen-off,
+            // which aborts these multi-minute sockets. Started here (user-triggered, app visible).
+            VideoDownloadService.start(appContext)
             activeJobs[id] = scope.launch {
                 try {
                     runCatching { performDownload(item, id) }
@@ -105,6 +109,7 @@ class VideoDownloadManager(
                     withContext(NonCancellable) { cleanupIfIncomplete(id) }
                     activeJobs.remove(id)
                     _progress.update { it - id }
+                    if (activeJobs.isEmpty()) VideoDownloadService.stop(appContext)
                 }
             }
         }
@@ -117,7 +122,7 @@ class VideoDownloadManager(
             runCatching { job.join() }
         }
         dao.findById(id)?.let { deleteQuietly(File(it.filePath)) }
-        deleteQuietly(File(downloadsDir, "$id.mp4"))
+        deleteQuietly(File(downloadsDir, "$id$FILE_EXT"))
         dao.deleteById(id)
     }
 
@@ -128,6 +133,7 @@ class VideoDownloadManager(
         }
         jobs.forEach { it.cancel() }
         jobs.forEach { runCatching { it.join() } }
+        VideoDownloadService.stop(appContext)
         _progress.value = emptyMap()
         dao.clear()
         withContext(Dispatchers.IO) {
@@ -142,8 +148,8 @@ class VideoDownloadManager(
         // Incomplete row first, so the Downloads UI lists the title while bytes stream in.
         dao.upsert(item.toDownloadedVideo(path = "", size = 0, qualityLabel = quality.label, complete = false))
         _progress.update { it + (id to VideoDownloadProgress(percent = 0, bytes = 0)) }
-        val target = File(downloadsDir, "$id.mp4")
-        val part = File(downloadsDir, "$id.mp4.part")
+        val target = File(downloadsDir, "$id$FILE_EXT")
+        val part = File(downloadsDir, "$id$FILE_EXT.part")
         val saved = streamToFile(url, part, id)
         if (saved && part.renameTo(target)) {
             dao.upsert(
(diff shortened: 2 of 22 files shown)


app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadManager.kt
```diff
@@ -0,0 +1,267 @@
...
+            activeJobs[id] = scope.launch {
+                try {
+                    runCatching { performDownload(item, id) }
+                        .onFailure { Log.w(TAG, "Video download failed for $id", it) }
+                } finally {
+                    // Also runs on success, where the row is complete and the cleanup no-ops.
+                    withContext(NonCancellable) { cleanupIfIncomplete(id) }
+                    activeJobs.remove(id)
+                    _progress.update { it - id }
+                    if (activeJobs.isEmpty()) VideoDownloadService.stop(appContext)
```
jaxkodex: **Race condition:** this `finally` block runs outside `synchronized(activeJobs)`, while `download()` calls `VideoDownloadService.start()` *before* inserting the new job into the map (both under the lock). Interleaving:

1. Thread B (new download): acquires lock, calls `start()` …
2. Thread A (finishing download): `activeJobs.remove(id)` → map momentarily empty → calls `stop()`
3. Thread B: `activeJobs[id] = launch { … }`

Result: the foreground service is stopped right after B started it, and B's multi-minute download runs without FGS protection — exactly the screen-off abort this service exists to prevent. Wrapping the `remove` + `isEmpty` + `stop()` sequence in `synchronized(activeJobs)` closes the window.

app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadManager.kt
```diff
@@ -0,0 +1,267 @@
...
+        if (completedPaths.containsKey(id)) return
+        synchronized(activeJobs) {
+            if (activeJobs.containsKey(id)) return
+            // Foreground service for the duration: Android freezes cached processes on screen-off,
+            // which aborts these multi-minute sockets. Started here (user-triggered, app visible).
+            VideoDownloadService.start(appContext)
+            activeJobs[id] = scope.launch {
+                try {
+                    runCatching { performDownload(item, id) }
+                        .onFailure { Log.w(TAG, "Video download failed for $id", it) }
```
jaxkodex: **Silent failure UX:** every failure path — HTTP error, low storage (`transcodeUrlOrNull` → null), not authenticated, mid-stream network drop — ends here with only a logcat line. From the user's perspective the button shows "Downloading…" and then quietly reverts to "Download" with no explanation. Consider surfacing failures (e.g. a `SharedFlow<String>` of error events the Detail/Downloads screens can toast/snackbar), especially for the low-storage case where retrying will never succeed.

app/src/main/java/pe/net/libre/mixtapehaven/data/jellyfin/JellyfinRepository.kt
```diff
@@ -252,67 +249,35 @@ class JellyfinRepository(
...
+        val client = api
+        val id = runCatching { UUID.fromString(itemId) }.getOrNull()
+        if (client == null || id == null) return null
+        return client.videosApi
+            .getVideoStreamUrl(
+                itemId = id,
+                container = "ts",
+                static = false,
+                // The server requires a media source; the default source id is the item id unhyphenated.
+                mediaSourceId = itemId.replace("-", ""),
```
jaxkodex, line 273: **Fragile assumption:** deriving `mediaSourceId` as the unhyphenated item id happens to hold for plain single-file library items, but it's an undocumented implementation detail — items with multiple versions/media sources (or `.strm`-backed items) have different source ids, and the transcode request will 400 or pick the wrong source. The robust path is a `PlaybackInfo` call (`mediaInfoApi.getPostedPlaybackInfo`) and using `mediaSources.first().id`, same as real clients do. If that's deliberately deferred, a TODO here noting the constraint would help the next reader.

app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadService.kt
```diff
@@ -0,0 +1,56 @@
...
+ * its own: [VideoDownloadManager] starts it with the first active download and stops it with the
+ * last, and the mandatory notification is the user's cue that a download is still running.
+ */
+class VideoDownloadService : Service() {
+
+    override fun onBind(intent: Intent?): IBinder? = null
+
+    override fun onStartCommand(intent: Intent?, flags: Int, startId: Int): Int {
+        createChannel()
+        startForeground(NOTIFICATION_ID, buildNotification())
```
jaxkodex: **Start/stop timing edge:** after `startForegroundService()`, the system requires this `startForeground()` call within ~5s. If `stop()` lands before `onStartCommand` runs (instant failure in `transcodeUrlOrNull` — low storage or unauthenticated — or an immediate cancel), the service can be stopped before ever calling `startForeground`, which on some API levels throws `ForegroundServiceDidNotStartInTimeException` / `RemoteServiceException` and crashes the app. Moving `startForeground()` into `onCreate()` shrinks the window to nearly zero, since `onCreate` runs as soon as the service is instantiated.

app/src/main/java/pe/net/libre/mixtapehaven/data/download/DownloadDatabase.kt
```diff
@@ -4,20 +4,38 @@ import android.content.Context
...
 import androidx.room.Room
 import androidx.room.RoomDatabase
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
 
-/** Room database holding the offline download library. */
-@Database(entities = [DownloadedTrack::class], version = 1, exportSchema = false)
+/** Room database holding the offline download library (audio tracks and videos). */
+@Database(entities = [DownloadedTrack::class, DownloadedVideo::class], version = 2, exportSchema = false)
 abstract class DownloadDatabase : RoomDatabase() {
```
jaxkodex, line 12: Now that the database has a real migration, consider flipping `exportSchema = true` (plus the `room.schemaLocation` ksp arg). That gives you (1) Room's schema JSON diffable in review, and (2) the ability to write a `MigrationTestHelper` instrumented test for `MIGRATION_1_2`, so future schema drift between this hand-written SQL and the `DownloadedVideo` entity is caught by CI instead of a crash-on-open in production. The live-install verification in the PR covers today's shape but won't protect v2→v3.

jaxkodex (commented):
## Review: manual video downloads with quality-capped transcodes

**Overview.** Adds user-triggered offline video downloads: a `DownloadedVideo` Room entity + DAO behind an additive v1→v2 migration, a `VideoDownloadManager` mirroring the audio manager (partial-file cleanup, free-space guard, in-flight rows), a minimal dataSync FGS to survive screen-off, quality-capped TS transcode URLs from `JellyfinRepository`, and UI in Detail / Downloads / Settings. Local-first playback reuses the step-1 `videoSourceResolver` seam with zero player changes.

**Overall this is solid.** The architecture mirrors the existing audio download path closely, the migration is correctly additive, cancellation/cleanup paths are thought through (NonCancellable cleanup, `.part` rename), and the PR description shows real on-device verification including the DB migration against a live v1 install. The refactors (extracting `JellyfinAudioMapping`, splitting `DownloadsScreen` into named composables) are behavior-preserving and improve readability.

### Correctness
- **Race on the FGS stop check** — the `finally` block's `activeJobs.isEmpty()` → `stop()` runs outside the lock that `download()` holds while starting the service. See inline comment; this can kill the FGS for a just-started download.
- **`mediaSourceId` derived by stripping hyphens** is an undocumented-shape assumption that breaks for multi-source/strm items. See inline comment.
- **FGS start/stop timing** — a stop landing before `onStartCommand` can trip the `startForegroundService` 5-second contract. See inline comment.

### UX / minor
- All failure modes (HTTP error, low storage, not authenticated) end with the Download button silently reverting — no user-visible error. See inline comment.
- The Settings "Wi-Fi only" toggle (`wifiOnly` remember state) is not consulted by `VideoDownloadManager`, so multi-GB downloads run on cellular regardless of the toggle. Fine if it's a known placeholder, but worth a TODO tying it to this path.
- Notification text in `VideoDownloadService` is hardcoded (consistent with the rest of the app's inline strings, so just noting it).

### Test coverage
- Good targeted unit tests for `VideoDownloadQuality.fromName` and the `DownloadedVideo` mapping (including the unknown-kind fallback). The migration itself is only manually verified — see inline comment about `exportSchema` enabling a `MigrationTestHelper` test.
- `VideoDownloadManager` itself is untested; its constructor already accepts injectable `scope`/`client`/DAO, so a test with MockWebServer covering cancel-mid-stream cleanup would be cheap and valuable.

### Security
- `?ApiKey=` on the transcode URL means the token lands in server access logs and any intermediary logging; already flagged as deferred to step 3 (header-based auth), so no action here.
- Local files under `filesDir` (app-private) and `android:exported="false"` on the service are correct.

No blocking issues — the inline comments are the concrete follow-ups, with the FGS stop race being the one I'd fix before merge.

app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadManager.kt
jaxkodex: Fixed in bdf46c6: the finishing job now does remove + isEmpty + stop() under the same `synchronized(activeJobs)` lock that `download()` holds while calling start() and inserting the job, so the interleaving described can no longer observe a momentarily-empty map.

app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadManager.kt
jaxkodex: Fixed in bdf46c6: `VideoDownloadManager` now emits one message per failure on a `SharedFlow` (distinct messages for low storage, unavailable URL, HTTP error, and generic failure — user cancels filtered out, including OkHttp's cancel-as-IOException shape), and the Video Detail screen collects it into a Toast.

app/src/main/java/pe/net/libre/mixtapehaven/data/jellyfin/JellyfinRepository.kt
jaxkodex: Deliberately deferred — documented the constraint at both call sites (download + HLS) with a pointer to `mediaInfoApi.getPostedPlaybackInfo` / `mediaSources.first().id` as the robust path. Will pick it up with the step-3 items. (Wrote 'Known limitation' rather than TODO because detekt's ForbiddenComment blocks TODO.)

app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadService.kt
jaxkodex: Fixed in bdf46c6: `startForeground()` moved into `onCreate()` (runs at instantiation, so a stop() landing before `onStartCommand` can no longer beat it) and now passes `FOREGROUND_SERVICE_TYPE_DATA_SYNC` explicitly as API 34+ requires. Re-verified on device: fresh download shows `isForeground=true types=0x00000001`, and an immediate cancel stops the service without a crash.

app/src/main/java/pe/net/libre/mixtapehaven/data/download/DownloadDatabase.kt
jaxkodex: Done in bdf46c6: `exportSchema = true` + `room.schemaLocation` ksp arg, with `schemas/…/2.json` committed so future migrations are diffable in review. Skipped the `MigrationTestHelper` instrumented test for now since CI doesn't run connectedAndroidTest — happy to add it if instrumented tests join the pipeline.

PR #150, opened 2026-07-19
feat: continue watching, next episode, and offline resume (video step 3)

jaxkodex:
Closes the four tier-1 gaps in the video feature, picked from a gap audit of steps 1 (#148) and 2 (#149). Subtitles and audio-track selection were deliberately deferred.

## Continue watching rail

Home gains a "Continue watching" rail between "On your device" and "Movies & shows", built to the `Continue Card` component in `happypath.pen`: a 16:9 still with a 3dp progress bar on its bottom edge, the title, and a `S2 E4 · 23 min left` meta line with an offline check when a local copy exists.

Tapping a card resumes straight in the player, skipping the detail screen. The rail merges the server's `getResumeItems` with the local progress table, so **it still works with no network**.

## Next episode

`selectPlayTarget` now prefers the server's Next Up episode. Jellyfin zeroes a watched episode's position, so the old "first in-progress episode" scan sent a user who had just finished E1 back to E1 instead of E2. The local scan remains as the offline fallback.

The player also auto-advances when an episode ends, and offers a skip pill. Both are keyed on **season position** rather than Next Up — autoplay should roll into the literal next episode, which is a different question from "where do I resume this series".

The pill is a bespoke control rather than `PlayerView`'s next button: the player holds a single `MediaItem` at a time (each episode negotiates its own stream), so the built-in one would render permanently disabled.

## Offline resume

Watch positions were server-only, so a downloaded title — precisely the one watched offline — reported into the void and restarted from zero on the next launch.

Positions now also land in a new `video_progress` table (migration 2 → 3, schema exported). On resume, whichever side was **written more recently** wins. Comparing by recency rather than taking the larger position matters: a deliberate rewind on one device would otherwise be undone by the other side's stale, larger value.

## PlaybackInfo negotiation

Transcode URLs assumed the media source id was the item id unhyphenated, which holds for plain single-file items but breaks for multi-version and `.strm`-backed ones. Both the HLS stream URL and the download URL now negotiate it via `mediaInfoApi.getPostedPlaybackInfo`, falling back to the old guess so a negotiation failure degrades rather than blocks playback.

## API shape change

`videoSourceResolver`, `videoStreamCandidates`, and `videoDownloadUrl` are now `suspend` — negotiating the media source id is a network call.

## Verification

Device-verified on a Pixel 8 against a live Jellyfin server:

| Check | Result |
|---|---|
| v2 → v3 migration on an existing DB | clean, no crash |
| Rail rendering | real stills, correct progress bars and time-left |
| Resume — movie | *Before Midnight* opened deep into the film, matching "22 min left" |
| Resume — episode | S3 E2 resumed mid-episode |
| Skip pill | showed S3 E3 while playing S3 E2; tapping advanced to E3, pill became S3 E4 |
| Stream negotiation | h264 decoder + audio focus confirmed in logcat |

Gates: `compileDebugKotlin`, `detekt`, and `assembleDebug` all clean. Unit tests **57 passing**, up from 42 — new coverage for `resolveResumePosition` (including the rewind case), `isFinished`, `mergeContinueWatching`, `formatTimeLeft`, and the Next Up branches of `selectPlayTarget`.

## Still open after this

Subtitles remain the largest gap, followed by audio-track selection, PiP, orientation lock, video search (`repository.search` is AUDIO-only), season grouping and paging, and the architectural point that video runs a raw ExoPlayer in a ViewModel — no MediaSession, hence no lock-screen/Bluetooth controls, background playback, or Android Auto.

### DIFF
--- a/app/src/main/java/pe/net/libre/mixtapehaven/data/jellyfin/JellyfinRepository.kt
+++ b/app/src/main/java/pe/net/libre/mixtapehaven/data/jellyfin/JellyfinRepository.kt
@@ -8,6 +8,7 @@ import org.jellyfin.sdk.api.client.extensions.audioApi
 import org.jellyfin.sdk.api.client.extensions.authenticateUserByName
 import org.jellyfin.sdk.api.client.extensions.dynamicHlsApi
 import org.jellyfin.sdk.api.client.extensions.itemsApi
+import org.jellyfin.sdk.api.client.extensions.mediaInfoApi
 import org.jellyfin.sdk.api.client.extensions.playStateApi
 import org.jellyfin.sdk.api.client.extensions.tvShowsApi
 import org.jellyfin.sdk.api.client.extensions.userApi
@@ -15,11 +16,16 @@ import org.jellyfin.sdk.api.client.extensions.userLibraryApi
 import org.jellyfin.sdk.api.client.extensions.videosApi
 import org.jellyfin.sdk.model.api.BaseItemKind
 import org.jellyfin.sdk.model.api.EncodingContext
+import org.jellyfin.sdk.model.api.ImageType
 import org.jellyfin.sdk.model.api.ItemFields
 import org.jellyfin.sdk.model.api.ItemSortBy
+import org.jellyfin.sdk.model.api.MediaType
 import org.jellyfin.sdk.model.api.PlayMethod
+import org.jellyfin.sdk.model.api.PlaybackInfoDto
 import org.jellyfin.sdk.model.api.SortOrder
 import org.jellyfin.sdk.model.api.request.GetItemsRequest
+import org.jellyfin.sdk.model.api.request.GetNextUpRequest
+import org.jellyfin.sdk.model.api.request.GetResumeItemsRequest
 import pe.net.libre.mixtapehaven.data.session.Session
 import pe.net.libre.mixtapehaven.data.session.SessionStore
 import pe.net.libre.mixtapehaven.model.Album
@@ -29,6 +35,12 @@ import java.util.UUID
 
 /** Lifecycle moment of video playback reported to the server via [JellyfinRepository.reportVideoPlayback]. */
 enum class VideoPlaybackEvent { STARTED, PROGRESS, STOPPED }
+
+/**
+ * Image types requested for Continue watching / Next Up rows. BACKDROP and THUMB are the wide art
+ * the 16:9 still wants; PRIMARY is the fallback (and is itself a 16:9 still for episodes).
+ */
+private val CONTINUE_IMAGE_TYPES = listOf(ImageType.BACKDROP, ImageType.THUMB, ImageType.PRIMARY)
 
 /** Wraps the Jellyfin Kotlin SDK: authentication, library browsing, search, and stream/image URLs. */
 class JellyfinRepository(
@@ -186,11 +198,34 @@ class JellyfinRepository(
     }
 
     /**
+     * The id of the media source the server picked for [itemId], via a PlaybackInfo negotiation.
+     *
+     * Transcode URLs need a real MediaSource id. For a plain single-file library item that happens
+     * to equal the item id unhyphenated, but multi-version items (a 4K and a 1080p cut) and
+     * .strm-backed ones have their own ids, and guessing sends the server the wrong source. Falls
+     * back to the guess when the call fails, so a negotiation error degrades to the old behaviour
+     * rather than blocking playback.
+     */
+    suspend fun mediaSourceId(itemId: String): String {
+        val fallback = itemId.replace("-", "")
+        val client = api
+        val id = runCatching { UUID.fromString(itemId) }.getOrNull()
+        if (client == null || id == null) return fallback
+        return runCatching {
+            val info by client.mediaInfoApi.getPostedPlaybackInfo(
+                itemId = id,
+                data = PlaybackInfoDto(userId = userId, enableDirectPlay = true, enableTranscoding = true),
+            )
+            info.mediaSources.firstOrNull()?.id
+        }.getOrNull() ?: fallback
+    }
+
+    /**
      * Ordered stream URL candidates for [itemId]: direct play of the original file first, then an
      * HLS transcode pinned to h264/aac for anything the device cannot play natively. The player
      * tries them in order and falls through on error.
      */
-    fun videoStreamCandidates(itemId: String): List<String> {
+    suspend fun videoStreamCandidates(itemId: String): List<String> {
         val client = api
         val id = runCatching { UUID.fromString(itemId) }.getOrNull()
         if (client == null || id == null) return emptyList()
@@ -200,13 +235,55 @@ class JellyfinRepository(
         val hls = client.dynamicHlsApi
             .getMasterHlsVideoPlaylistUrl(
                 itemId = id,
-                // Known limitation: assumes the default single media source (see videoDownloadUrl).
-                mediaSourceId = itemId.replace("-", ""),
+                mediaSourceId = mediaSourceId(itemId),
                 videoCodec = "h264",
                 audioCodec = "aac",
             )
             .withApiKey(client)
         return listOf(direct, hls)
+    }
+
+    /**
+     * The next episode to watch in [seriesId] per the server's Next Up logic, or null when the
+     * series is finished or unwatched. Unlike scanning episodes for a resume position this is
+     * correct after an episode is watched to completion (Jellyfin zeroes its position, so a local
+     * scan would send the user back to the episode they just finished).
+     */
+    suspend fun nextUpEpisode(seriesId: String): VideoItem? {
+        val client = requireApi()
+        val id = runCatching { UUID.fromString(seriesId) }.getOrNull() ?: return null
+        val result by client.tvShowsApi.getNextUp(
+            GetNextUpRequest(
+                userId = userId,
+                seriesId = id,
+                fields = listOf(ItemFields.OVERVIEW),
+                limit = 1,
+                // Include the partially-watched episode itself, not just the one after it.
+                enableResumable = true,
+                enableImages = true,
+                enableImageTypes = CONTINUE_IMAGE_TYPES,
+            ),
+        )
+        return result.items.orEmpty().firstOrNull()?.toVideoItem(client)
+    }
+
+    /** The server's Continue Watching list: partially-watched movies and episodes, most recent first. */
+    suspend fun continueWatching(limit: Int = 12): List<VideoItem> {
+        val client = requireApi()
+        val result by client.itemsApi.getResumeItems(
+            GetResumeItemsRequest(
+                userId = userId,
+                limit = limit,
+                mediaTypes = listOf(MediaType.VIDEO),
+                includeItemTypes = listOf(BaseItemKind.MOVIE, BaseItemKind.EPISODE),
+                fields = listOf(ItemFields.OVERVIEW),
+                // Resume items come back without image tags unless asked for, which would leave
+                // every Continue watching still on its flat colour fallback.
+                enableImages = true,
+                enableImageTypes = CONTINUE_IMAGE_TYPES,
+            ),
+        )
+        return result.items.orEmpty().map { it.toVideoItem(client) }
     }
 
     /**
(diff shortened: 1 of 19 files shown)


app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerViewModel.kt
```diff
@@ -84,19 +101,86 @@ class VideoPlayerViewModel(
...
     }
 
+    /** Load [id], resume it, and begin playing. Used for the initial item and each autoplay hop. */
+    private suspend fun startItem(id: String) {
+        val resolved = progressStore.resolvePlayback(id)
+        val item = resolved.item
+        if (item == null) {
+            _error.value = "Could not load this title"
+            return
+        }
```
jaxkodex, line 119: **Blocking: offline first play of a downloaded title is now broken.**

`resolvePlayback` sources the item from `repository.videoItem(id)` (network — null offline) or the local `video_progress` row. A title that has been **downloaded but never played** has neither offline, so this returns early with "Could not load this title" — the exact scenario the offline feature exists for.

The previous code only used `videoItem()` for a resume position and played regardless, so this is a regression against step 2.

`downloaded_videos` already carries title/kind/seriesName/seasonEpisodeLabel/posterUrl/artColor. Suggest a third fallback in `VideoProgressStore.resolvePlayback` (inject `VideoDownloadDao`), or at minimum synthesize a bare `VideoItem(id = id, ...)` so playback proceeds and reporting stays best-effort.

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerViewModel.kt
```diff
@@ -84,19 +101,86 @@ class VideoPlayerViewModel(
...
+        _nowPlaying.value = item
+        candidates = resolveSources(id)
+        candidateIndex = 0
+        reachedReady = false
+        if (candidates.isEmpty()) {
+            _error.value = "No playable source for this title"
+            return
```
jaxkodex, line 126: _upNext_ isn't reset here, so an autoplay hop that bails at this branch leaves the pill pointing at the *previous* episode's next. Clearing `_upNext.value = null` at the top of `startItem`, alongside `candidateIndex` and `reachedReady`, also removes the window where the pill shows a stale value while `loadUpNext` is in flight.

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerViewModel.kt
```diff
@@ -84,19 +101,86 @@ class VideoPlayerViewModel(
...
+     *
+     * Deliberately positional rather than the server's Next Up: autoplay should roll into the
+     * literal next episode, whereas Next Up answers the different question of where to resume a
+     * series (see [VideoDetailViewModel]).
+     */
+    private suspend fun loadUpNext(item: VideoItem): VideoItem? {
+        val seriesId = item.seriesId?.takeIf { item.kind == VideoKind.EPISODE } ?: return null
+        val episodes = runCatching { repository.seriesEpisodes(seriesId) }.getOrNull().orEmpty()
+        val index = episodes.indexOfFirst { it.id == item.id }
+        return if (index >= 0) episodes.getOrNull(index + 1) else null
```
jaxkodex, line 146: Autoplay correctness now silently depends on `getEpisodes` returning season/episode order — `seriesEpisodes` passes no explicit sort, it just relies on the Jellyfin default. `parentIndexNumber`/`indexNumber` are already mapped for the label, so sorting defensively here (or in `seriesEpisodes`) would make the dependency explicit rather than a load-bearing assumption about server behaviour.

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/home/HomeViewModel.kt
```diff
@@ -152,3 +189,23 @@ class HomeViewModel(
...
+ */
+internal fun mergeContinueWatching(
+    local: List<VideoItem>,
+    server: List<VideoItem>,
+): List<VideoItem> = (local + server)
+    .groupBy { it.id }
+    .map { (_, records) -> records.maxBy { it.lastPlayedAtMs } }
+    .filter { it.resumePositionMs > 0 }
+    .sortedByDescending { it.lastPlayedAtMs }
+    .take(CONTINUE_WATCHING_LIMIT)
```
jaxkodex, line 218: **A finished title comes back into the rail.**

`record()` deletes the local row once `isFinished`, but `serverContinueWatching` is a snapshot only refreshed by `load()` — which doesn't re-run when you pop back from the player onto a surviving `HomeViewModel`. The server copy still has `resumePositionMs > 0`, and with the local row gone there's no fresher record to beat it, so the episode you just finished re-surfaces here.

Options: keep a tombstone set of locally-finished ids and subtract it in the merge, or re-`load()` when Home resumes.

app/src/main/java/pe/net/libre/mixtapehaven/data/playback/VideoProgressStore.kt
```diff
@@ -0,0 +1,102 @@
...
+        } else {
+            dao.upsert(item.toProgress(positionMs, runtimeMs, now()))
+        }
+        repository.reportVideoPlayback(item.id, positionMs, event, paused = paused, transcoding = transcoding)
+    }
+
+    /** Forget the local position for [itemId] (e.g. when its download is deleted). */
+    suspend fun clear(itemId: String) = dao.deleteById(itemId)
+
+    suspend fun clearAll() = dao.clear()
```
jaxkodex, line 83: Both `clear` and `clearAll` are currently unreachable — nothing calls either.

Two consequences:
- `clear`'s KDoc says "e.g. when its download is deleted", but download deletion doesn't call it, so a deleted download keeps its progress row.
- More importantly, sign-out (`MixtapeNavHost.kt:68`) doesn't touch `video_progress`. After switching Jellyfin accounts or servers, the new user sees the previous user's Continue watching rail — titles, stills and positions — until they play something. Small privacy leak worth closing by wiring `clearAll()` into sign-out.

app/src/main/java/pe/net/libre/mixtapehaven/data/jellyfin/JellyfinRepository.kt
```diff
@@ -185,12 +197,35 @@ class JellyfinRepository(
...
+        val fallback = itemId.replace("-", "")
+        val client = api
+        val id = runCatching { UUID.fromString(itemId) }.getOrNull()
+        if (client == null || id == null) return fallback
+        return runCatching {
+            val info by client.mediaInfoApi.getPostedPlaybackInfo(
+                itemId = id,
+                data = PlaybackInfoDto(userId = userId, enableDirectPlay = true, enableTranscoding = true),
+            )
+            info.mediaSources.firstOrNull()?.id
```
jaxkodex, line 232: `mediaSources.firstOrNull()` takes whatever the server happens to list first, so for the multi-version case cited in the KDoc above (a 4K and a 1080p cut) this is still arbitrary — it properly fixes `.strm`-backed items, but not that one. Worth either narrowing the KDoc claim or, in `videoDownloadUrl`, matching the source against the requested `maxHeight`.

Also worth noting: this is an extra POST per stream resolve *and* per download start, and `getPostedPlaybackInfo` opens a server-side play session. Fine at this scale, but it now sits on the playback critical path.

jaxkodex (commented):
## Review

Four related additions on top of video steps 1–2: the Continue watching rail, Next Up–driven Play target plus in-player autoplay, offline resume via the v2→v3 migration, and PlaybackInfo negotiation for the media source id.

The design reasoning is unusually well documented — the recency-over-max-position rule, the positional-vs-NextUp split for autoplay, and the bespoke pill are all justified in KDoc rather than left implicit. Unit coverage on the new pure functions is genuinely good; `testDebugUnitTest` is clean locally.

**One blocking issue** (offline first play of a downloaded title now errors out), plus two worth folding in before merge. Details inline.

### Summary

| # | Severity | Issue |
|---|---|---|
| 1 | high | `startItem` hard-fails when neither the server nor a local progress row can supply the item — i.e. a downloaded, never-played title, offline |
| 2 | medium | A finished title re-surfaces in the rail from the stale server snapshot until Home reloads |
| 3 | medium | `clear`/`clearAll` are dead code; sign-out leaves the previous user's watch history in the rail |
| 4 | low | `mediaSourceId` takes `mediaSources.first()`, which is still arbitrary for multi-version items |
| 5 | low | `_upNext` goes stale when a hop fails at the `candidates.isEmpty()` branch |
| 6 | low | `loadUpNext` silently depends on `getEpisodes` default ordering |

### Smaller notes

- `ContinueCard`'s clickable sets no `role`/`onClickLabel`, while `BackButton`/`NextEpisodePill` in the same PR do set `role = Role.Button`. Inconsistent for TalkBack.
- `formatTimeLeft` computes `remainingMs` before guarding `runtimeMs <= 0`. Harmless, reads better reordered.
- `orZero()` is a greedy name for a file-private `Long` extension; `durationOrZero()` would read better at the call sites.
- The `HomeSectionActions` extraction (7-param composable → 2) is a good call and consistent with the file's style.

jaxkodex:
Thanks — several of these were real bugs. All addressed in `5dca3af`.

### Fixed

| # | Issue | Resolution |
|---|---|---|
| 1 | Offline first play of a downloaded title fails | `resolvePlayback` gains the `downloaded_videos` row as a third fallback, so a downloaded-but-never-played title resolves offline |
| — | `seriesId` missing from `VideoProgress` (@gemini-code-assist) | Added to the entity, both mappers, and the migration SQL |
| 2 | Finished title re-surfaces from the stale server snapshot | Finished titles are kept as position-0 **tombstones** instead of deleted, and subtracted in `mergeContinueWatching` |
| 3 | `clear`/`clearAll` dead; sign-out leaks watch history | `clearAll()` wired into sign-out |
| 5 | `_upNext` goes stale on a failed hop | Reset at the top of `startItem`, which also closes the window where the pill shows a stale value while `loadUpNext` runs |
| 6 | `loadUpNext` depends on `getEpisodes` default ordering | `seriesEpisodes` now sorts explicitly by `parentIndexNumber`/`indexNumber` |

The `seriesId` catch is the one I'd have been most annoyed to ship: it made autoplay and the skip pill die **specifically offline**, which is the one path where the local row is load-bearing — and the online device test could never have caught it.

### Handled differently

**#3, `clear(itemId)`** — removed rather than wired to download deletion. The KDoc was the thing that was wrong: deleting a download shouldn't forget a position the server still holds, since the title remains in the library. `clearAll` stays, now used.

**#4, `mediaSources.first()`** — narrowed the claim rather than implementing quality matching. The KDoc now says plainly that it fixes `.strm`-backed items and is *not* a considered choice between versions of a multi-version item, and notes the extra POST on the playback path plus the server-side play session. Picking by requested quality means matching `maxHeight` against each source's streams; worth doing, but it belongs with proper multi-version support rather than bolted on here.

**`formatTimeLeft` i18n** (@gemini-code-assist) — deferred deliberately. It's a fair point, but user-facing strings are literals throughout this codebase's Compose layer (`"Continue watching"`, `"Movies & shows"`, `"On your device"`); extracting only this one would be inconsistent. Worth a single pass over the video UI rather than a one-off.

**`maxBy` → `maxByOrNull`** — applied, though for a different reason than stated: `maxBy` is not deprecated on this Kotlin version (it compiles warning-free), but the non-null variant avoids a throw-on-empty foot-gun if the grouping ever changes.

Also picked up the smaller notes: `Role.Button` + `onClickLabel` on the Continue card, the `formatTimeLeft` guard order, and `orZero` → `durationOrZero`.

### Verification

Re-verified on a Pixel 8 **in airplane mode** (device reported `Active default network: none`):

- Continue watching renders from the local table while "Recently added" correctly shows the offline state
- *Before Sunset* — downloaded, with no `video_progress` row, i.e. exactly the never-played case — plays from local storage

Tests **59 passing** (up from 57), with new coverage for the `seriesId` round-trip and the finished-title suppression. `detekt` and `assembleDebug` clean.

One note for reviewers: `MIGRATION_2_3` was amended in place rather than adding a 3→4. That's safe because this branch has never shipped, but any device that already installed an earlier build of *this branch* needs the app data cleared.

PR #151, opened 2026-07-19
feat: video search, library browse, and seasons (video step 4)

jaxkodex:
Step 4 of the video feature. Steps 1–3 gave us streaming (#148), downloads (#149), and continue watching (#150) — but video was reachable only through a 20-item Home rail and was invisible to search. This makes the library browsable.

## Discovery

- **Search covers video.** `JellyfinRepository.search()` filtered `AUDIO` only; there is now a parallel `searchVideos()` over movies, series, and episodes, run concurrently with the track query so a library with no video doesn't add a sequential round trip.
- **Library screen** behind a new "See all" on the Home rail: paged poster grid with kind (All/Movies/Series), genre, and sort facets.
  - Paging is offset-based, which is only safe because every sort option pins a *total* order by falling back to `SORT_NAME`. Without that tiebreak, items with no year/rating come back in arbitrary server order and paging silently duplicates and drops them.
  - A failed "load more" keeps what's on screen; only a fresh load shows the error state.

## Watch state

- `VideoItem` gains `played`, `unplayedCount`, `genres`, `communityRating`, `officialRating`, `seasonNumber`.
- Posters show a tick when finished, or an unwatched-episode count for a part-watched series.
- **Seasons:** series detail groups episodes into season chips and opens on the season holding Next Up — a viewer mid-season-4 shouldn't scroll past three finished seasons. Season 0 is bucketed as "Specials" and pushed last (it sorts *before* season 1 numerically, so a series would otherwise open on its specials).
- Episode rows carry a watched tick; the detail header gains content rating, community score, and genres.

## Notes

`PosterCard` no longer hardcodes 120.dp — the Home rail pins a width, the grid sizes from its cell.

## Testing

- New unit tests: `SeasonGroupingTest` (7), `InitialSeasonTest` (4). Full suite green.
- Lint clean of new issues.
- Verified on device against a live Jellyfin server: kind filters, genre chips, sort menu, paging past the first page, season switching, watched/unplayed badges, and video search results.


app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoLibraryViewModel.kt
```diff
@@ -0,0 +1,129 @@
...
+                    filter = facets.filter,
+                    genre = facets.genre,
+                    sort = facets.sort,
+                    startIndex = startIndex,
+                )
+            }
+            _state.update { state ->
+                page.fold(
+                    onSuccess = { result ->
+                        val items = if (append) state.items + result.items else result.items
```
jaxkodex: Offset paging over a live library can return an item already in `items` — a title added (or a re-sort under `RECENTLY_ADDED`) between page 1 and page 2 shifts everything by one, and the boundary item comes back twice. `LazyVerticalGrid` is keyed by `it.id`, and a duplicate key is a hard crash (`IllegalArgumentException: Key "…" was already used`), not a cosmetic dupe.

The `SORT_NAME` tiebreak fixes *arbitrary* ordering but not a *changing* result set, so it doesn't cover this.

```kotlin
val items = if (append) {
    val seen = state.items.mapTo(HashSet()) { it.id }
    state.items + result.items.filterNot { it.id in seen }
} else {
    result.items
}
```

Note this also means `items.size` drifts below the true offset; `endReached` still terminates correctly since `totalCount` is refreshed each page, but a dropped duplicate makes the next `startIndex` re-read one row. Acceptable for the cost.

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoLibraryScreen.kt
```diff
@@ -0,0 +1,308 @@
...
+    modifier: Modifier = Modifier,
+) {
+    val viewModel = appViewModel { VideoLibraryViewModel(it.repository) }
+    val state by viewModel.state.collectAsState()
+    val gridState = rememberLazyGridState()
+
+    // Paging is driven off the last visible index rather than an onClick "Load more" so the grid
+    // fills ahead of the user; distinctUntilChanged keeps every scroll pixel from re-asking.
+    LaunchedEffect(gridState, viewModel) {
+        snapshotFlow { gridState.layoutInfo.visibleItemsInfo.lastOrNull()?.index ?: 0 }
```
jaxkodex, line 86: `gridState` isn't reset when a facet changes. Scroll deep into All/Recently added, then tap Movies: `items` is replaced from index 0 but the grid keeps its offset, so the user lands mid-list in a set they never scrolled — and if the new set is shorter, `LazyGrid` clamps to the end, so the screen opens on the last row with nothing above it looking loaded.

Worth having the ViewModel expose a facet generation/reset signal and scrolling on it:

```kotlin
LaunchedEffect(state.filter, state.genre, state.sort) {
    gridState.scrollToItem(0)
}
```

Separately: the prefetch effect reads `state.items.size` inside the `collect` lambda. That resolves against the composable's `State` object so it does see fresh values — just flagging that it's subtle enough to be worth a word in the existing comment, since `distinctUntilChanged` is keyed only on the index and the size is the other half of the condition.

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/search/SearchViewModel.kt
```diff
@@ -25,8 +28,11 @@ class SearchViewModel(
     data class UiState(
         val query: String = "",
         val results: List<Track> = emptyList(),
+        val videos: List<VideoItem> = emptyList(),
         val loading: Boolean = false,
-    )
+    ) {
+        val isEmpty: Boolean get() = results.isEmpty() && videos.isEmpty()
```
jaxkodex, line 36: `isEmpty` has no callers — `SearchScreen` gates on `state.results.isNotEmpty()` / `state.videos.isNotEmpty()` directly and the count label sums the two sizes. Either drop it, or use it for the "no results" empty state that the screen currently doesn't render (a query with zero hits shows only "0 results").

app/src/main/java/pe/net/libre/mixtapehaven/data/jellyfin/JellyfinRepository.kt
```diff
@@ -42,6 +44,49 @@ enum class VideoPlaybackEvent { STARTED, PROGRESS, STOPPED }
...
+ */
+private fun VideoSort.sortBy(): List<ItemSortBy> = when (this) {
+    VideoSort.RECENTLY_ADDED -> listOf(ItemSortBy.DATE_CREATED, ItemSortBy.SORT_NAME)
+    VideoSort.TITLE -> listOf(ItemSortBy.SORT_NAME)
+    VideoSort.YEAR -> listOf(ItemSortBy.PRODUCTION_YEAR, ItemSortBy.SORT_NAME)
+    VideoSort.RATING -> listOf(ItemSortBy.COMMUNITY_RATING, ItemSortBy.SORT_NAME)
+}
+
+/** Only titles sort descending; alphabetical order is the one case where ascending is wanted. */
+private fun VideoSort.sortOrder(): List<SortOrder> = when (this) {
```
jaxkodex: `sortBy()` returns two keys for YEAR/RATING/RECENTLY_ADDED but `sortOrder()` returns one. Jellyfin pairs them positionally and defaults the unmatched key to ascending, which happens to be exactly what you want for the `SORT_NAME` tiebreak — but the whole paging-stability argument in the KDoc above rests on that unstated default.

Since the tiebreak is load-bearing, make it explicit:

```kotlin
VideoSort.YEAR -> listOf(SortOrder.DESCENDING, SortOrder.ASCENDING)
```

Also worth noting `RATING` descending puts unrated titles (null community rating) at the *end* on most Jellyfin versions, which is probably the intent — but it's the one sort where the tiebreak does the heavy lifting for a large slice of a typical library.

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/search/SearchScreen.kt
```diff
@@ -69,10 +78,25 @@ fun SearchScreen(onBack: () -> Unit, onOpenNowPlaying: () -> Unit, modifier: Mod
...
+        // Video leads: a title match is a stronger intent signal than the song matches that
+        // usually accompany it (soundtracks share the film's name).
+        if (state.videos.isNotEmpty()) {
+            SectionHeader("Movies & shows")
+            LazyRow(horizontalArrangement = Arrangement.spacedBy(12.dp)) {
+                items(state.videos, key = { it.id }) { video ->
+                    PosterCard(
+                        video = video,
+                        onClick = { onOpenVideo(video.id) },
+                        modifier = Modifier.width(120.dp),
```
jaxkodex: The PR notes say `PosterCard` no longer hardcodes `120.dp` — but this reintroduces the literal at a second call site, so the rail width now lives in two places (`HomeScreen.POSTER_RAIL_WIDTH` and here) and can drift.

Since both are the same "poster rail" affordance, promote the constant next to `PosterCard` in `Components.kt` and have both rails use it.

app/src/main/java/pe/net/libre/mixtapehaven/data/jellyfin/JellyfinVideoMapping.kt
```diff
@@ -50,9 +50,41 @@ internal fun BaseItemDto.toVideoItem(client: ApiClient): VideoItem {
...
+ * Episodes grouped into seasons in ascending season order, with [SPECIALS_LABEL] last.
+ *
+ * Season 0 is Jellyfin's specials bucket and sorts *before* season 1 numerically, so it is pushed
+ * to the end explicitly — a series should open on its first real season, not its specials.
+ */
+internal fun groupBySeason(episodes: List<VideoItem>): List<Pair<String, List<VideoItem>>> =
+    episodes
+        .groupBy { it.seasonNumber ?: 0 }
+        .toList()
+        .sortedBy { (season, _) -> if (season <= 0) Int.MAX_VALUE else season }
```
jaxkodex: `groupBy { it.seasonNumber ?: 0 }` keys on the raw number, but `seasonLabel` collapses everything `<= 0` to `"Specials"`. A negative `parentIndexNumber` (rare, but Jellyfin doesn't forbid it) therefore yields two distinct groups that both render as `"Specials"` — and `SeasonSelector` uses `items(seasons, key = { it })`, so duplicate labels crash the `LazyRow` the same way duplicate ids would.

Grouping on the bucket rather than the number removes the class of bug entirely:

```kotlin
episodes
    .groupBy { it.seasonNumber?.takeIf { n -> n > 0 } }
    .toList()
    .sortedBy { (season, _) -> season ?: Int.MAX_VALUE }
    .map { (season, items) -> seasonLabel(season) to items }
```

`SeasonGroupingTest` already covers null-and-zero sharing a bucket; a `-1` case would pin this down.

jaxkodex (commented):
## Review — video search, library browse, and seasons

Solid, well-documented step 4. The comments explain *why* (the SORT_NAME tiebreak, specials-last, keep-on-screen for failed load-more) rather than restating the code, and the new pure functions (`groupBySeason`, `initialSeason`) are cleanly extracted and unit-tested. Concurrent track/video search is the right call.

Few things worth addressing before merge — details inline.

**Higher priority**
- Offset paging can append a duplicate id when the library changes mid-scroll; `LazyVerticalGrid` with `key = { it.id }` throws on a duplicate key. (`VideoLibraryViewModel.kt`)
- Facet changes reset `items` but not the grid scroll position. (`VideoLibraryScreen.kt`)

**Lower priority**
- `UiState.isEmpty` is dead code.
- `sortOrder()` returns one order for two sort keys — works, but only by relying on Jellyfin's positional pairing default.
- `120.dp` re-hardcoded at the Search call site, undoing part of the `PosterCard` cleanup.
- A negative season number produces two "Specials" groups → duplicate key in the season `LazyRow`.

**Test coverage:** `SeasonGroupingTest`/`InitialSeasonTest` are good, but the trickiest new logic — `VideoLibraryViewModel`'s paging/facet-reset/append-vs-error state machine — has none. A fake repository plus `runTest` would cover append, end-reached, and the failed-load-more-keeps-items path cheaply.

**Security/perf:** nothing concerning. No new URL or credential handling; `enableImages = false` on the genres call and the concurrent search are both good. `VIDEO_FIELDS` adds OVERVIEW to the Home rail payload, which is unused there but negligible at 20 items.

jaxkodex:
Pushed `065d3c0`. CI green locally: **83 tests, detekt clean, lint clean.**

## CI failure (detekt, 6 violations)

The real signal was `JellyfinRepository` hitting 21 functions — it had quietly become a god-object over auth + music + video. Split the browse/search queries into `JellyfinVideoLibrary` behind the `VideoLibrarySource` interface (which also gave the ViewModel tests their seam). `moviesAndShows()` is deleted outright: after this PR it was *exactly* `videoLibrary(limit = 20).items`.

The rest were threshold trips from grouping-related params: extracted `PosterCard`/`WatchBadge` to their own file, pulled out `libraryDestinations()` and `SearchBar()`, and introduced `VideoNavActions` / `SeasonUi` / `DownloadUi`.

## @jaxkodex — all addressed

- **Duplicate id on append** — fixed via `appendPage()`. You were right that the `SORT_NAME` tiebreak only makes the *ordering* total and does nothing for a *changing* result set; I've said so explicitly in the KDoc, including the offset-drift note.
- **Grid scroll not reset on facet change** — `UiState.facetGeneration`, scrolled to top by a `LaunchedEffect`. Verified on device: scrolled deep into All, tapped Series, landed at row 1.
- **`isEmpty` dead code** — took your second option; it now backs a real "No matches in your library" state, since a zero-hit query previously rendered only "0 results".
- **`sortOrder()` positional pairing** — spelled both entries out. Agreed the tiebreak shouldn't rest on an unstated server default.
- **`120.dp` re-hardcoded** — `POSTER_RAIL_WIDTH` now lives next to `PosterCard`; both rails use it.
- **Negative season → duplicate "Specials"** — grouping is on the bucket (`seasonNumber?.takeIf { it > 0 }`), so it's structurally impossible now. Added the `-1` case plus a test asserting labels are unique, since that's the actual crash precondition.
- **Prefetch `state.items.size` subtlety** — documented in the existing comment.
- **ViewModel tests** — `VideoLibraryViewModelTest`, 11 tests over a fake `VideoLibrarySource` with `runTest`: append, dedupe-on-append, end-reached (asserts no further request), facet reset, genre toggle-off, first-load error, and the failed-load-more-keeps-items path. Needed `kotlinx-coroutines-test`, added to the catalog.

## @gemini-code-assist — 3 of 7 valid

**Applied:** the `CancellationException` catch (genuinely load-bearing — a facet change cancels the in-flight page and the failure branch was clobbering the replacement job's state) and both 48.dp touch targets. Your `groupBySeason` normalization was also correct; I used the bucket-key form instead, which removes the class of bug rather than the instance.

**Not applied — the three nullability suggestions are incorrect.** `result.items` and `result.totalRecordCount` are *not* nullable in jellyfin-model 1.8.11:

```
BaseItemDtoQueryResult:
  items: List<BaseItemDto>   // non-null
  totalRecordCount: Int      // non-null primitive
```

`.orEmpty()` on a non-null `List` and `?: 0` on a non-null `Int` would be no-ops. The proof is that the call sites compile: `result.items.map { … }` and `totalCount = result.totalRecordCount` would both be compile errors against nullable types. The `.orEmpty()` elsewhere in the file is legacy from an older SDK, not evidence of nullability.

PR #152, opened 2026-07-19
feat: robust video downloads via WorkManager (video step 5)

jaxkodex:
## Summary

Video downloads currently run as plain coroutines held alive by a bare `dataSync` foreground service: process death loses the transfer, a network blip fails a GB-sized download with a toast, N taps run N concurrent server transcodes, and there is no way to keep them off mobile data. This PR moves the pipeline onto **WorkManager** and adds a **"Download over Wi-Fi only"** setting (default ON).

- **Queue survives process death** — `VideoDownloadWorker` re-runs automatically after force-stop/reboot; a global mutex serializes transfers so the Jellyfin server only ever runs one transcode for us.
- **Retry with backoff** — transient failures (IO, 5xx) retry up to 5 attempts (30s exponential); 4xx/low-storage fail immediately. Failed rows get a retry button on Downloads and a "Download failed · Retry" state on the detail screen.
- **Wi-Fi-only, default ON** — `UNMETERED` WorkManager constraint for video (rows show "Waiting for Wi-Fi"); audio auto-download does a best-effort metered-network skip.
- **DB v3→v4** — additive `status` column (`QUEUED/RUNNING/FAILED/COMPLETE`) + startup reconciliation for rows orphaned by pruned work or process death.
- `VideoDownloadService` deleted; WorkManager''s `SystemForegroundService` gets the `dataSync` type via manifest merge-override, and its notification now shows live byte counts.

**Design note:** retries restart from byte 0 — the source is a live server-side transcode whose bytes are not stable across ffmpeg runs, so Range-resuming a `.part` would splice two streams into a corrupt file.

## Test plan

- [x] `detekt`, unit tests, lint, `assembleDebug` green
- [x] Device: DB migration opens existing library intact
- [x] Device: queue 2 episodes → sequential drain (second shows "Waiting…")
- [x] Device: force-stop mid-download → relaunch auto-restarts worker, completes
- [x] Device: Wi-Fi off mid-download → "Waiting for Wi-Fi"; Wi-Fi on → auto-resume
- [x] Device: airplane-mode playback of a completed download (regression)
- [x] Device: Settings toggle persists; default ON


app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadWorker.kt
```diff
@@ -0,0 +1,126 @@
...
+            .setSmallIcon(R.drawable.ic_launcher_foreground)
+            .setContentTitle("Downloading video")
+            .setContentText(text)
+            .setOngoing(true)
+            .build()
+
+    companion object {
+        private const val TAG = "VideoDownloadWorker"
+        private const val CHANNEL_ID = "video_downloads"
+        private const val NOTIFICATION_ID = 20
```
jaxkodex: **Shared notification ID across concurrent workers (main concern).** With the mutex, two workers routinely coexist: A streams while B waits, and B has already called `setForeground` with this same ID 20. When A finishes, WorkManager's `SystemForegroundService` cleans up A's notification/foreground association by ID — which can tear down B's notification or its foreground status mid-wait; WorkManager's internal notificationId→workspec bookkeeping doesn't handle duplicate IDs well.

Suggest deriving the ID per work item (e.g. `NOTIFICATION_ID_BASE + id.hashCode()`), or keeping one ID deliberately but routing all notification updates through a single owner.

app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadManager.kt
```diff
@@ -163,96 +193,196 @@ class VideoDownloadManager(
...
-        val saved = streamToFile(url, part, id)
-        if (saved && part.renameTo(target)) {
+        return try {
+            if (shouldStartDownload(usableSpaceBytes(), MIN_FREE_BYTES)) {
+                streamAtQuality(row, settingsStore.videoQuality.first(), part)
+            } else {
+                Log.w(TAG, "Skipping video download of ${row.id}: storage low")
+                failPermanently(row.id, "Not enough free space to download")
+            }
+        } catch (failure: IOException) {
```
jaxkodex: **Uncaught non-`IOException` leaves a zombie row until next launch.** Only `IOException` is caught here. Anything else (Room error, unexpected repository exception) propagates out of `doWork`, WorkManager records `Result.failure()` with no retry — but the `finally` block demotes the row to QUEUED, so the UI shows "Waiting…" forever with no live work behind it, until `reconcileQueue()` on the next app start flips it to FAILED.

Catching `Exception` (re-throwing `CancellationException`) and routing unexpected ones to `failPermanently` would close the gap.

app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadWorker.kt
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    override suspend fun getForegroundInfo(): ForegroundInfo = foregroundInfo()
+
+    private suspend fun toResult(outcome: VideoDownloadOutcome, id: String): Result = when (outcome) {
+        VideoDownloadOutcome.COMPLETED, VideoDownloadOutcome.SKIPPED -> Result.success()
+        VideoDownloadOutcome.FAILED -> Result.failure()
+        VideoDownloadOutcome.RETRY ->
+            // runAttemptCount is 0-based, so this allows MAX_ATTEMPTS runs in total.
+            if (runAttemptCount < MAX_ATTEMPTS - 1) {
```
jaxkodex: **Verify `runAttemptCount` vs constraint stops.** Workers stopped by constraint loss (Wi-Fi drops mid-stream) can increment `runAttemptCount` without ever returning `Result.retry()`. On flaky Wi-Fi, a later genuine transient failure may find `runAttemptCount` already ≥ 4 and `markFailed` a download that only ever lost its network.

Worth an on-device check: toggle Wi-Fi a few times mid-download, then force one real retry. If confirmed, track your own attempt counter (e.g. in the row or in work progress data) instead of relying on `runAttemptCount`.

app/src/main/AndroidManifest.xml
```diff
@@ -1,50 +1,52 @@
...
+                <action android:name="android.intent.action.MAIN" />
+
+                <category android:name="android.intent.category.LAUNCHER" />
+            </intent-filter>
+        </activity>
+
+        <!-- WorkManager runs VideoDownloadWorker as a foreground data sync on API 34+; its
+             service must declare the type or startForeground throws. Merged, not replaced. -->
+        <service
+            android:name="androidx.work.impl.foreground.SystemForegroundService"
```
jaxkodex, line 43: **`dataSync` FGS time budget on Android 15+.** With targetSdk 36, `dataSync` foreground services are capped at ~6 hours per 24h. A queue of several GB downloads on slow Wi-Fi can hit this — the system stops the service and the workers with it. WorkManager will re-run them, but attempts churn (compounding the `runAttemptCount` concern). Not a blocker; worth a comment here and a known-limitation note in the PR.

app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadManager.kt
```diff
@@ -163,96 +193,196 @@ class VideoDownloadManager(
...
+     * pruned, data cleared) becomes FAILED so the UI offers retry instead of a stuck row, and a
+     * RUNNING row whose worker died with the process drops back to QUEUED until its re-run streams.
      */
-    private suspend fun transcodeUrlOrNull(id: String, quality: VideoDownloadQuality): String? {
-        val alreadySaved = dao.findById(id)?.complete == true
-        val lowStorage = !shouldStartDownload(usableSpaceBytes(), MIN_FREE_BYTES)
-        if (lowStorage) {
-            Log.w(TAG, "Skipping video download of $id: storage low")
-            _errors.tryEmit("Not enough free space to download")
+    private suspend fun reconcileQueue() {
```
jaxkodex, line 350: **Startup race:** a `download()` landing between `incompleteIds()` and the WorkManager query can get its fresh QUEUED row marked FAILED here. It self-heals when the worker runs (it doesn't check status), but the user sees a FAILED flicker. A guard like skipping QUEUED rows enqueued in the last few seconds would remove it. Minor.

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/downloads/DownloadsViewModel.kt
```diff
@@ -1,119 +1,143 @@
...
+        videoProgress: Map<String, VideoDownloadProgress>,
+        wifiOnly: Boolean,
+    ): UiState {
+        val audioBytes = completedTracks.sumOf { it.sizeBytes }
+        val videoBytes = videoRows.filter { it.complete }.sumOf { it.sizeBytes }
+        val free = videoDownloadManager.usableSpaceBytes()
+        // The queue is only ever gated on network when the Wi-Fi-only constraint is set and the
+        // current network would not satisfy it — that's when "Waiting…" deserves the reason.
+        val waitingLabel =
+            if (wifiOnly && videoDownloadManager.isNetworkMetered()) "Waiting for Wi-Fi" else "Waiting…"
```
jaxkodex: **Label can go stale:** `isNetworkMetered()` is only re-evaluated when one of the combined flows emits, so switching networks doesn't refresh "Waiting for Wi-Fi" ↔ "Waiting…" until something else changes. A `ConnectivityManager` callback flow as a combine source would fix it; fine to defer.

app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadManager.kt
```diff
@@ -264,39 +394,49 @@ class VideoDownloadManager(
...
+
+        /** Tag on every video download work request, for bulk cancellation. */
+        const val WORK_TAG = "video-download"
+
+        /** Unique WorkManager name for the download of [id], shared by enqueue/cancel/reconcile. */
+        fun uniqueWorkName(id: String): String = "video-download-$id"
     }
 }
+
+private fun deleteQuietly(file: File) {
```
jaxkodex, line 450: Nit: moving these to file-level private functions re-hardcodes the `"VideoDownloadManager"` tag. Keeping them in the companion (or passing the tag) avoids the duplicate string.

app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadManager.kt
```diff
@@ -100,61 +134,57 @@ class VideoDownloadManager(
...
-    /** Start downloading [item] (a movie or episode). No-op if already saved or in flight. */
+    /** True when the active network bills by the byte, i.e. queued work is waiting for Wi-Fi. */
+    fun isNetworkMetered(): Boolean =
+        appContext.getSystemService(ConnectivityManager::class.java)?.isActiveNetworkMetered == true
+
+    /**
+     * Queue [item] (a movie or episode) for download. No-op if already saved or queued; re-taps
+     * on a failed row re-queue it. The transfer starts when its worker gets network + its turn.
+     */
     fun download(item: VideoItem) {
```
jaxkodex, line 147: **Edge case:** `download()` on an already-RUNNING row upserts it back to QUEUED (resetting the quality label mid-stream). The current UI can't trigger this — the detail button shows Cancel while in flight — but a `status == RUNNING` guard would make the API safe by construction.

app/src/test/java/pe/net/libre/mixtapehaven/data/download/DownloadedVideoMappingTest.kt
```diff
@@ -1,41 +1,54 @@
...
+
+    @Test
+    fun `download status round-trips by name`() {
+        VideoDownloadStatus.entries.forEach { status ->
+            assertEquals(status, VideoDownloadStatus.fromName(status.name))
+        }
+    }
+
+    @Test
+    fun `an unknown persisted status degrades to failed so the UI offers retry`() {
```
jaxkodex, line 51: These enum tests are good, but the risky surfaces are untested:

- **Migration 3→4** via Room's `MigrationTestHelper` (schemas are exported precisely to enable this)
- **`VideoDownloadWorker`** via `androidx.work:work-testing` (`TestListenableWorkerBuilder`) — at minimum the `toResult` mapping and exhausted-retries → `markFailed`
- `reconcileQueue` transitions (orphaned row → FAILED, dead RUNNING → QUEUED)

The device test plan covers the happy paths well; these would lock the logic in for future refactors.

jaxkodex (commented):
Review of the WorkManager migration (video step 5). Overall: solid, well-reasoned PR — the worker/manager split via `VideoDownloadOutcome` is clean, every outcome writes its own status before returning, the v3→v4 migration is additive and exported, and the no-resume design note (transcode bytes unstable across ffmpeg runs) documents a real constraint in the right place.

I'd fix two things before merge (shared notification ID, uncaught non-IOException zombie row), verify one on-device (`runAttemptCount` vs constraint stops), and the rest can be follow-ups — details inline.

Housekeeping: the diff GitHub shows (~+2900/−2284) is inflated by CRLF→LF rewrites of whole files; the real change is ~+844/−225. Consider a `.gitattributes` with `* text=auto eol=lf` in a separate commit so future reviews show real diffs.

app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadWorker.kt
jaxkodex: Fixed in 00960bf: notification ids are now per-item (2000-2999 from the item-id hash), and `setForeground` moved inside the transcode mutex — so only the actively streaming worker holds a foreground notification, and A finishing can no longer tear down B's association. Waiting workers stay covered by the active worker's FGS keeping the shared process unfrozen.

app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadManager.kt
jaxkodex: Fixed in 00960bf: `attemptRun` now catches `Exception` (rethrowing `CancellationException`), routes `IOException` to the transient path and everything else to `failPermanently` — no more zombie QUEUED row.

app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadWorker.kt
jaxkodex: Sidestepped rather than verified: 00960bf drops `runAttemptCount` entirely and tracks the retry budget in a new `attempts` column (migration 4->5), incremented only on genuine transient failures — constraint stops surface as `CancellationException` and never touch it. Manual retry resets the budget (`dao.requeue`). The policy itself is unit-tested (`outcomeAfterTransient`).

app/src/main/AndroidManifest.xml
jaxkodex: Documented in the manifest in 00960bf. With the own-counter change the attempt-churn compounding is gone: a system stop spends no retry budget, WorkManager just re-runs the worker.

app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadManager.kt
jaxkodex: Agreed it self-heals; deferring per your note. The clean fix is skipping QUEUED rows younger than a few seconds, which wants an enqueuedAt column — will fold into a follow-up if the flicker ever shows up in practice.

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/downloads/DownloadsViewModel.kt
jaxkodex: Deferring per your note — a ConnectivityManager callback flow as a combine source is the right shape; noted as follow-up.

app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadManager.kt
jaxkodex: Fixed in 00960bf: TAG is now a file-level `private const` shared by the class and the file-level helpers.

app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadManager.kt
jaxkodex: Fixed in 00960bf: `download()` now returns early when the row is RUNNING, so a re-tap can't reset the row under the active worker.

app/src/test/java/pe/net/libre/mixtapehaven/data/download/DownloadedVideoMappingTest.kt
jaxkodex: Added in 00960bf: instrumented migration tests over the exported schemas (v3->v5: complete row keeps data + gains defaults, incomplete row surfaces as FAILED, sibling tables survive) plus unit tests for the retry-budget policy. Worker-level tests via work-testing and reconcile transitions still pending — the worker's mapping shrank to a trivial when() now that the budget lives in the manager, so the remaining gap is reconcile; noting as follow-up.

app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadWorker.kt
```diff
@@ -0,0 +1,126 @@
...
+            coroutineScope {
+                val notifier = launch { publishProgress(id) }
+                try {
+                    toResult(manager.runDownload(id), id)
+                } finally {
+                    notifier.cancel()
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
jaxkodex: Done in 00960bf — setForeground now happens after acquiring the mutex, combined with per-item notification ids so concurrent workers can't clobber each other.

app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadManager.kt
```diff
@@ -163,96 +193,196 @@ class VideoDownloadManager(
...
+                // Only cancellation (worker stopped, network constraint lost) exits with the row
+                // still RUNNING — every outcome writes its own status first. Drop it back to
+                // QUEUED so the UI shows "Waiting…" instead of a dead spinner until the re-run.
+                if (after != null && !after.complete && after.status == VideoDownloadStatus.RUNNING.name) {
+                    dao.updateStatus(id, VideoDownloadStatus.QUEUED.name)
+                }
+                _progress.update { it - id }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
jaxkodex: Done in 00960bf — runDownload is wrapped in withContext(Dispatchers.IO).

app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadManager.kt
```diff
@@ -163,96 +193,196 @@ class VideoDownloadManager(
...
                 if (!response.isSuccessful) {
                     Log.w(TAG, "Video download of $id got HTTP ${response.code}")
-                    _errors.tryEmit("Download failed (server error ${response.code})")
-                    false
+                    StreamResult.HttpError(response.code)
                 } else {
                     copyBody(response.body, part, id)
+                    StreamResult.SAVED
                 }
             }
```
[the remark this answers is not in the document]
jaxkodex: Not applicable on OkHttp 5.x (this project uses 5.4.0): `Response.body` is non-nullable there — a null check would be flagged as always-false by the compiler.

app/src/main/java/pe/net/libre/mixtapehaven/data/download/DownloadManager.kt
```diff
@@ -1,249 +1,267 @@
...
+        if (lowStorage) Log.w(TAG, "Skipping download of $id: storage low")
+        return if (alreadySaved || lowStorage) null else repository.audioStreamUrl(id)
+    }
+
+    /** Stream [url] into [part], emitting progress for [track]. Returns false if cancelled mid-stream. */
+    private suspend fun streamToFile(url: String, part: File, track: Track): Boolean {
+        val request = Request.Builder().url(url).build()
+        return client.newCall(request).execute().use { response ->
+            if (!response.isSuccessful) false else copyToFile(response.body, part, track)
+        }
```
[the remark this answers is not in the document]
jaxkodex: Not applicable on OkHttp 5.x (this project uses 5.4.0): `Response.body` is non-nullable there.

app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadManager.kt
```diff
@@ -163,96 +193,196 @@ class VideoDownloadManager(
...
-        val alreadySaved = dao.findById(id)?.complete == true
-        val lowStorage = !shouldStartDownload(usableSpaceBytes(), MIN_FREE_BYTES)
-        if (lowStorage) {
-            Log.w(TAG, "Skipping video download of $id: storage low")
-            _errors.tryEmit("Not enough free space to download")
+    private suspend fun reconcileQueue() {
+        for (id in dao.incompleteIds()) {
+            val alive = withContext(Dispatchers.IO) {
+                runCatching { workManager.getWorkInfosForUniqueWork(uniqueWorkName(id)).get() }.getOrNull()
+            }?.any { !it.state.isFinished }
```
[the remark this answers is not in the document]
jaxkodex: Done in spirit in 00960bf: `androidx.work.await` is for `Operation`, not `ListenableFuture`, so reconcileQueue uses `getWorkInfosForUniqueWorkFlow(...).first()` — properly suspending, no blocking .get().

app/src/main/java/pe/net/libre/mixtapehaven/data/download/VideoDownloadManager.kt
```diff
@@ -163,96 +193,196 @@ class VideoDownloadManager(
...
+    }
+
+    /** 4xx means the request itself is wrong and will stay wrong; anything else is worth a retry. */
+    private suspend fun httpOutcome(id: String, code: Int): VideoDownloadOutcome =
+        if (code in CLIENT_ERRORS) {
+            failPermanently(id, "Download failed (server error $code)")
+        } else {
+            dao.updateStatus(id, VideoDownloadStatus.QUEUED.name)
+            VideoDownloadOutcome.RETRY
         }
```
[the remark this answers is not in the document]
jaxkodex: @
Fixed in 00960bf - message is now "Download failed (error <code>)".
@

PR #161, opened 2026-08-04
Add buffering indicator to video player

jaxkodex:
## Summary
Adds a buffering state indicator to the video player screen that displays a loading spinner while the player is resolving streams, buffering content, or preparing fallback transcodes.

## Key Changes
- **VideoPlayerViewModel:** Introduced a new `buffering` StateFlow that tracks when the player surface has nothing to show
  - Set to `true` when entering `STATE_BUFFERING` or during network operations (stream resolution, source loading)
  - Set to `false` when reaching `STATE_READY`, `STATE_ENDED`, or when errors occur
  - Properly managed during transcode fallback attempts to indicate the swap is another wait, not a resume
  
- **VideoPlayerScreen:** Added a `CircularProgressIndicator` that displays centered on the player when buffering
  - Uses the app's `Accent` color for visual consistency
  - Suppressed when an error message is displayed (error message better explains the black frame)
  - 48dp size for appropriate visibility

## Implementation Details
The buffering flag is raised proactively during network operations (item resolution, source loading) rather than waiting for the player's `STATE_BUFFERING` event, ensuring the indicator appears immediately during these initial setup phases. The flag is properly cleared in all error paths to prevent the spinner from displaying indefinitely.

https://claude.ai/code/session_01CZ4vHvauXk5hkNvWHoxkYW

### DIFF
--- a/app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerScreen.kt
+++ b/app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerScreen.kt
@@ -27,6 +27,8 @@ import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.semantics.Role
+import androidx.compose.ui.semantics.contentDescription
+import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.viewinterop.AndroidView
@@ -121,7 +123,11 @@ fun VideoPlayerScreen(
 private fun BufferingIndicator(modifier: Modifier = Modifier) {
     CircularProgressIndicator(
         color = Accent,
-        modifier = modifier.size(48.dp),
+        // Without this the load is silent to TalkBack — the black frame with no announcement is
+        // the exact state this indicator exists to explain.
+        modifier = modifier
+            .size(48.dp)
+            .semantics { contentDescription = "Loading video" },
     )
 }
 
--- a/app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerScreen.kt
+++ b/app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerScreen.kt
@@ -13,6 +13,7 @@ import androidx.compose.foundation.shape.CircleShape
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.SkipNext
 import androidx.compose.material.icons.outlined.ArrowBack
+import androidx.compose.material3.CircularProgressIndicator
 import androidx.compose.material3.Icon
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
@@ -35,6 +36,7 @@ import androidx.lifecycle.compose.LocalLifecycleOwner
 import androidx.media3.common.util.UnstableApi
 import androidx.media3.ui.PlayerView
 import pe.net.libre.mixtapehaven.di.appViewModel
+import pe.net.libre.mixtapehaven.ui.theme.Accent
 import pe.net.libre.mixtapehaven.ui.theme.TextPrimary
 
 /** Full-screen video playback surface using media3's [PlayerView] for transport controls. */
@@ -59,6 +61,7 @@ fun VideoPlayerScreen(
     }
     val error by viewModel.error.collectAsState()
     val upNext by viewModel.upNext.collectAsState()
+    val buffering by viewModel.buffering.collectAsState()
 
     // No background video service exists, so pause when the screen stops (Home button, lock);
     // otherwise audio would keep playing invisibly and the progress loop would churn the radio.
@@ -84,6 +87,15 @@ fun VideoPlayerScreen(
             onRelease = { it.player = null },
             modifier = Modifier.fillMaxSize(),
         )
+
+        // Suppressed once an error is up: the message explains the black frame better than a
+        // spinner that would never resolve.
+        if (buffering && error == null) {
+            CircularProgressIndicator(
+                color = Accent,
+                modifier = Modifier.align(Alignment.Center).size(48.dp),
+            )
+        }
 
         error?.let { message ->
             Text(
--- a/app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerScreen.kt
+++ b/app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerScreen.kt
@@ -91,10 +91,7 @@ fun VideoPlayerScreen(
         // Suppressed once an error is up: the message explains the black frame better than a
         // spinner that would never resolve.
         if (buffering && error == null) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerViewModel.kt
```diff
@@ -111,9 +130,13 @@ class VideoPlayerViewModel(
         // Cleared up front so a hop that fails below cannot leave the pill pointing at the
         // previous episode's successor, and so it never shows a stale value while loadUpNext runs.
         _upNext.value = null
+        // Resolving the item and its sources are network hops with the player still idle, so the
+        // flag is raised here rather than waiting for STATE_BUFFERING.
+        _buffering.value = true
```
jaxkodex, line 147: A stale error suppresses the spinner on the next hop. `_buffering` is raised here, but `_error` isn't cleared until line 152 — after both network calls. So if the current title already failed (`"No playable source for this title"`, or a playback error with no candidates left) and the user then taps the next-episode pill, the screen holds the *previous* item's error text for the whole resolve window, and the screen's `buffering && error == null` gate keeps the spinner hidden. It reads as frozen.

Clearing it alongside `_upNext` fixes it:

```kotlin
_upNext.value = null
_error.value = null
_buffering.value = true
```

The existing `_error.value = null` on line 152 then becomes redundant.

---

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerViewModel.kt
```diff
@@ -111,9 +130,13 @@ class VideoPlayerViewModel(
...
         val resolved = progressStore.resolvePlayback(id)
         val item = resolved.item
         if (item == null) {
+            _buffering.value = false
```
jaxkodex, line 160: Nit / defensive: the "flag always gets cleared" invariant currently depends on the callees never throwing. `resolvePlayback` and the default `videoSourceResolver` are both `runCatching`-guarded today, so this isn't reachable — but `resolveSources` is an injected lambda, and where a throw used to leave a black frame it would now leave a spinner turning forever with no message.

Wrapping the body of `startItem` in a `try/catch` that clears `_buffering` and sets a generic error would keep the invariant local to this function instead of resting on what the callees happen to do.

---

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerViewModel.kt
```diff
@@ -76,8 +84,16 @@ class VideoPlayerViewModel(
     private val listener = object : Player.Listener {
         override fun onPlaybackStateChanged(playbackState: Int) {
             when (playbackState) {
-                Player.STATE_READY -> reachedReady = true
-                Player.STATE_ENDED -> onPlaybackEnded()
+                Player.STATE_BUFFERING -> _buffering.value = true
```
jaxkodex, line 88: Worth knowing before you ship: `STATE_BUFFERING` also fires on seeks inside an already-buffered file, so scrubbing will flash the spinner over the transport controls. Common enough behavior that it may be fine as-is — if it looks noisy on device, the usual remedy is only surfacing the indicator after ~200–300ms of continuous buffering rather than gating the state itself.

---

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerScreen.kt
```diff
@@ -107,6 +116,15 @@ fun VideoPlayerScreen(
     }
 }
 
+/** Spinner shown over the black frame while a stream is being resolved or buffered. */
+@Composable
+private fun BufferingIndicator(modifier: Modifier = Modifier) {
+    CircularProgressIndicator(
+        color = Accent,
+        modifier = modifier.size(48.dp),
+    )
```
jaxkodex, line 131: The indicator carries no semantics, so TalkBack announces nothing during load — a screen-reader user gets silence over a black frame, which is exactly the state this PR is meant to explain. Adding a description would close that:

```kotlin
CircularProgressIndicator(
    color = Accent,
    modifier = modifier
        .size(48.dp)
        .semantics { contentDescription = "Loading video" },
)
```

Everything else in the file (`BackButton`, `NextEpisodePill`) already labels itself, so this would stay consistent with the surrounding code.

---

jaxkodex (commented):
Reviewed the buffering indicator. Overall this is sound and idiomatic for the file — the call to raise the flag proactively during the network hops rather than waiting for `STATE_BUFFERING` is right, since the player sits in `STATE_IDLE` through `resolvePlayback`/`resolveSources` and would otherwise show a black frame with no feedback. Starting `_buffering` at `true` matches `init` launching straight into `startItem`, the transcode-fallback handling correctly keeps the indicator up across the swap, and the `STATE_IDLE` comment checks out (`stop()` is never called; `release()` only happens in `onCleared`). No perf or security surface touched.

One change worth making before merge — the stale-error case on line 135 — plus three optional notes inline.

**Test coverage:** none added, and Sonar shows 0% on new code. The repo does unit-test ViewModels (`VideoLibraryViewModelTest`), but `VideoPlayerViewModel` builds a real `ExoPlayer` in its constructor, so it isn't reachable without Robolectric. Reasonable to leave for now — the buffering flag is a pure state machine and would be the natural first thing to cover if that VM ever gets a testable seam.

---

jaxkodex:
Addressed in `1bd3492` — three of the four taken, one deferred.

**Stale error (line 135)** — confirmed and fixed; `_error` now clears at the top alongside `_upNext`, and the later assignment is gone. Worth noting the reachable path is narrower than the comment suggests: after either early-return error, `_upNext` is already null and no pill is rendered, so `playNext` can't be invoked. The one way in is a *terminal playback error* — `startItem` completes, sets `_upNext`, then all candidates are exhausted in `onPlayerError`. The pill stays on screen, and tapping it reproduces exactly the frozen-looking window you described.

**Throw guard (line 139)** — taken, and it's less hypothetical than "nit": `resolvePlayback` calls `dao.findById` (Room) with no `runCatching` at `VideoProgressStore.kt:44`, and `videoStreamCandidates` leaves both URL-builder calls unguarded — only the `UUID` parse and `mediaSourceId` are wrapped. So the path exists today without swapping the injected resolver. The load proper moved to `prepareItem` and `startItem` now owns the flag across every exit via `runCatching`, rethrowing `CancellationException` so scope shutdown still unwinds rather than surfacing as a load failure.

**Semantics (screen line 125)** — taken as written.

**Seek flash (line 87)** — deferred, not dismissed. Agreed it's real, but the fix is a timing threshold, and picking 200ms vs 300ms without watching a scrub on a device is guesswork; a debounce that's slightly too long also delays the indicator in the case this PR exists to fix. The APK artifact from this run is the way to settle it. Happy to add the delay in a follow-up if it looks noisy — say the word.

**Coverage** — agreed with your read, and it matches what we found separately: the buffering flag is a pure state machine, but `VideoPlayerViewModel` builds its `ExoPlayer` in the constructor, so it's unreachable without either Robolectric (which would also need `media3-test-utils-robolectric` to drive real state transitions) or a player seam. The seam is the better trade and would make this the natural first thing to cover.

Verified locally: `detekt`, `testDebugUnitTest`, and `lintDebug` all pass.

---

PR #162, opened 2026-08-09
fix: open the app from the now playing notification

jaxkodex:
MediaSession was built without a session activity, so media3 generated the
playback notification with no content intent — tapping the notification or the
system Now Playing tile did nothing.

Set a session activity that launches MainActivity with an extra requesting the
Now Playing route, and make MainActivity singleTask so a tap while the app is
already running brings the existing task forward (delivered via onNewIntent)
instead of restarting it. The requested route is dropped while signed out or
when already on Now Playing.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01Ks2BLHxFTPUUJRyjKwLrxy

### DIFF
--- a/app/src/main/java/pe/net/libre/mixtapehaven/MainActivity.kt
+++ b/app/src/main/java/pe/net/libre/mixtapehaven/MainActivity.kt
@@ -17,6 +17,7 @@ import kotlinx.coroutines.flow.asStateFlow
 import pe.net.libre.mixtapehaven.di.appViewModel
 import pe.net.libre.mixtapehaven.ui.navigation.AppStateViewModel
 import pe.net.libre.mixtapehaven.ui.navigation.MixtapeNavHost
+import pe.net.libre.mixtapehaven.ui.navigation.NoPendingRoute
 import pe.net.libre.mixtapehaven.ui.navigation.Routes
 import pe.net.libre.mixtapehaven.ui.theme.Bg
 import pe.net.libre.mixtapehaven.ui.theme.MixtapeTheme
@@ -25,14 +26,18 @@ class MainActivity : ComponentActivity() {
 
     /** Route requested by the intent that opened (or re-opened) the app; null once consumed. */
     private val _pendingRoute = MutableStateFlow<String?>(null)
+    private val pendingRoute: StateFlow<String?> = _pendingRoute.asStateFlow()
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         enableEdgeToEdge()
-        readPendingRoute(intent)
+        // Only on a fresh launch. A non-null savedInstanceState means the activity is being
+        // recreated (rotation, or Recents after a process kill), and the launch intent the system
+        // replays still carries the extra — honouring it would yank the user back to Now Playing.
+        if (savedInstanceState == null) readPendingRoute(intent)
         setContent {
             MixtapeApp(
-                pendingRoute = _pendingRoute.asStateFlow(),
+                pendingRoute = pendingRoute,
                 onRouteConsumed = { _pendingRoute.value = null },
             )
         }
@@ -50,6 +55,9 @@ class MainActivity : ComponentActivity() {
 
     private fun readPendingRoute(intent: Intent) {
         if (intent.getBooleanExtra(EXTRA_OPEN_NOW_PLAYING, false)) {
+            // Consume it: [setIntent] retains this intent, so leaving the extra in place would
+            // replay the request the next time the activity is recreated from it.
+            intent.removeExtra(EXTRA_OPEN_NOW_PLAYING)
             _pendingRoute.value = Routes.NOW_PLAYING
         }
     }
@@ -62,7 +70,7 @@ class MainActivity : ComponentActivity() {
 
 @Composable
 fun MixtapeApp(
-    pendingRoute: StateFlow<String?> = MutableStateFlow(null),
+    pendingRoute: StateFlow<String?> = NoPendingRoute,
     onRouteConsumed: () -> Unit = {},
 ) {
     val appState = appViewModel { AppStateViewModel(it.repository) }
--- a/app/src/main/java/pe/net/libre/mixtapehaven/MainActivity.kt
+++ b/app/src/main/java/pe/net/libre/mixtapehaven/MainActivity.kt
@@ -54,18 +54,24 @@ class MainActivity : ComponentActivity() {
     }
 
     private fun readPendingRoute(intent: Intent) {
-        if (intent.getBooleanExtra(EXTRA_OPEN_NOW_PLAYING, false)) {
-            // Consume it: [setIntent] retains this intent, so leaving the extra in place would
-            // replay the request the next time the activity is recreated from it.
-            intent.removeExtra(EXTRA_OPEN_NOW_PLAYING)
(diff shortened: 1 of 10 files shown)


jaxkodex:
@claude review this PR

app/src/main/java/pe/net/libre/mixtapehaven/MainActivity.kt
```diff
@@ -10,28 +11,72 @@ import androidx.compose.runtime.Composable
...
+    private val _pendingRoute = MutableStateFlow<String?>(null)
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         enableEdgeToEdge()
-        setContent { MixtapeApp() }
+        readPendingRoute(intent)
+        setContent {
+            MixtapeApp(
+                pendingRoute = _pendingRoute.asStateFlow(),
```
jaxkodex: **This breaks the build.** Calling the `asStateFlow()` flow operator inside the `setContent` composition trips Compose's `FlowOperatorInvokedInComposition` lint error, and the `build` job on this PR is already red because of it:

```
app/src/main/java/pe/net/libre/mixtapehaven/MainActivity.kt:35: Error: Flow operator functions should not be invoked within composition [FlowOperatorInvokedInComposition from androidx.compose.runtime]
                pendingRoute = _pendingRoute.asStateFlow(),
> Task :app:lintDebug FAILED
> Lint found errors in the project; aborting build.
```

Beyond CI, the underlying complaint is real: any recomposition of this lambda mints a new `StateFlow` instance, which resets the `collectAsState()` in `MixtapeNavHost`.

`MutableStateFlow` already *is* a `StateFlow`, and the parameter type is read-only, so passing the field directly fixes it with no wrapper:

```suggestion
                pendingRoute = _pendingRoute,
```

(Alternatively, hoist `private val pendingRoute: StateFlow<String?> = _pendingRoute.asStateFlow()` to a field and pass that.)

---

app/src/main/java/pe/net/libre/mixtapehaven/MainActivity.kt
```diff
@@ -10,28 +11,72 @@ import androidx.compose.runtime.Composable
...
+        super.onNewIntent(intent)
+        setIntent(intent)
+        readPendingRoute(intent)
+    }
+
+    private fun readPendingRoute(intent: Intent) {
+        if (intent.getBooleanExtra(EXTRA_OPEN_NOW_PLAYING, false)) {
+            _pendingRoute.value = Routes.NOW_PLAYING
+        }
+    }
```
jaxkodex, line 58: The extra is read but never removed from the intent, and `onNewIntent` persists that intent with `setIntent(intent)`. `MainActivity` declares no `android:configChanges`, so it is destroyed and recreated on every configuration change — and `onCreate` re-reads the *same* intent, so the request fires again.

Concrete repro:
1. Tap the playback notification → Now Playing.
2. Press back → Home.
3. Rotate the device.

`onCreate` sees `EXTRA_OPEN_NOW_PLAYING` still `true` → `_pendingRoute = NOW_PLAYING`; `rememberNavController` restores `currentDestination = home`, so `resolveDeepLink` returns `now_playing` and the user is yanked back to Now Playing. Same for font-size/theme changes, multi-window resize, and "Don't keep activities".

Worse case: if the app was *cold-started* from the notification, that intent becomes the task's base intent, so returning from Recents after a background process kill also reopens Now Playing instead of restoring the screen the user left.

Consume the extra when you read it:

```suggestion
    private fun readPendingRoute(intent: Intent) {
        if (intent.getBooleanExtra(EXTRA_OPEN_NOW_PLAYING, false)) {
            // Consume it: the intent is retained (setIntent) and replayed to onCreate on every
            // activity recreation, which would re-navigate to Now Playing on each rotation.
            intent.removeExtra(EXTRA_OPEN_NOW_PLAYING)
            _pendingRoute.value = Routes.NOW_PLAYING
        }
    }
```

---

app/src/main/java/pe/net/libre/mixtapehaven/ui/navigation/MixtapeNavHost.kt
```diff
@@ -24,11 +29,27 @@ import pe.net.libre.mixtapehaven.ui.screens.video.VideoLibraryScreen
 import pe.net.libre.mixtapehaven.ui.screens.video.VideoPlayerScreen
 
 @Composable
-fun MixtapeNavHost(startDestination: String, modifier: Modifier = Modifier) {
+fun MixtapeNavHost(
+    startDestination: String,
+    modifier: Modifier = Modifier,
+    pendingRoute: StateFlow<String?> = MutableStateFlow(null),
```
jaxkodex: Minor, but the same defect class lint flagged on `MainActivity.kt:35` — it just isn't detected here because it's a constructor rather than a flow operator. Composable default expressions are re-evaluated on every non-skipped recomposition, so any caller that omits `pendingRoute` (previews, Compose UI tests, a future call site) gets a fresh `StateFlow` identity each time, tearing down and restarting the `collectAsState()` collection on line 42.

```suggestion
    pendingRoute: StateFlow<String?> = remember { MutableStateFlow(null) },
```

(Or drop the defaults entirely — there's exactly one production caller.)

---

jaxkodex (commented):
Reviewed the diff for correctness. The overall approach (session activity → `singleTask` + `onNewIntent` → pending-route flow, with a pure `resolveDeepLink` covered by unit tests) is sound. Three issues, one of which is why CI is red:

1. `MainActivity.kt:35` — `asStateFlow()` in composition fails `./gradlew lint` (`FlowOperatorInvokedInComposition`), which is the current `build` failure.
2. `MainActivity.kt:51` — the `EXTRA_OPEN_NOW_PLAYING` extra is never removed, so every activity recreation (rotation, theme change, Recents restore after a process kill) replays the deep link and yanks the user back to Now Playing.
3. `MixtapeNavHost.kt:35` — default `MutableStateFlow(null)` allocated per recomposition (minor; same class as #1).

Details inline.

---

jaxkodex:
All three addressed in `2d49ca3`. `./gradlew lintDebug detekt testDebugUnitTest` is green locally.

**1. `asStateFlow()` in composition** — fixed via the hoisted-field alternative rather than passing `_pendingRoute` directly, so the type exposed to the composable stays read-only:

```kotlin
private val _pendingRoute = MutableStateFlow<String?>(null)
private val pendingRoute: StateFlow<String?> = _pendingRoute.asStateFlow()
```

**2. Extra never consumed** — real bug, thanks. Took the suggested `removeExtra`, plus a guard in `onCreate`:

```kotlin
if (savedInstanceState == null) readPendingRoute(intent)
```

`removeExtra` alone covers the in-process cases (rotation re-delivers the same mutated `Intent` instance), but not the cold-restore case you flagged: after a background process kill, the system replays the task's *persisted* base intent, which still carries the extra. The `savedInstanceState == null` check catches that — a fresh launch is the only time the extra should be honoured, and a real notification tap always produces one.

**3. Per-recomposition default** — used a shared instance instead of `remember`, so the identity is stable across call sites and not just within one composition:

```kotlin
val NoPendingRoute: StateFlow<String?> = MutableStateFlow(null)
```

Still unverified: the actual notification tap. There's no emulator in this session, so the `setSessionActivity` → `onNewIntent` → navigate path has only been reasoned through, not run on a device.

---

PR #163, opened 2026-08-09
Mark Continue watching titles that can't play, and say why

jaxkodex:
## The bug

With the server out of reach, the Continue watching rail listed every recently watched title identically, downloaded or not. Tapping one that had no saved copy navigated into the player, which then sat on a black frame while each stream candidate failed its own connection attempt — reading as a tap that did nothing.

The rail is fed by the local Room progress table, which is exactly why it survives without a server; but `videoSourceResolver` only serves a local file when a completed download exists, and otherwise hands back the two Jellyfin stream URLs.

"Out of reach" is deliberately not "no network": a Jellyfin published on the LAN only, or behind a VPN that is currently down, is unreachable from a perfectly healthy mobile connection.

## What changed

**`NetworkMonitor`** (new) — connectivity over `ConnectivityManager`. Requires `NET_CAPABILITY_INTERNET` but *not* `NET_CAPABILITY_VALIDATED`, since a LAN with no internet route still serves a self-hosted server.

**`ServerAvailability`** (new) — the verdict the UI actually needs, folding together:
- connectivity — a fast, certain no;
- `report(success)` from requests the app already makes — Home's own load reports in, costing nothing;
- `check()` — `/System/Ping` with a 3s budget when neither has a fresh answer. Pings are serialized behind a mutex and a success is trusted for 30s, so a burst of taps costs one request and a reachable server adds no latency. Regaining a network confirms rather than assumes; losing one is conclusive without asking.

**Home** — `UiState.serverReachable` drives the rail. A title with no completed download renders dimmed, with a cloud-off "Not downloaded" line in place of the time left. Tapping it goes through `HomeViewModel.resumeVideo`, which confirms with a ping before refusing (so a server that came back plays on the first tap, not after a manual reload) and otherwise sends a snackbar naming the real problem: "connect to watch it" with no network, "your server isn't reachable — check your VPN or network" with one.

**Player** — the same check before preparing, so any other entry point (autoplay into an undownloaded next episode, a deep link) fails fast with a message instead of a black frame. When every candidate errors out, it re-checks and swaps in the server message — that is what a server disappearing mid-playback looks like. A saved copy never pings at all.

## Testing

`./gradlew testDebugUnitTest` and `./gradlew lintDebug` both pass. New unit coverage: `ServerAvailabilityTest` (8 — including the online-but-unreachable case, ping caching, shared concurrent pings, and a server coming back), `ContinueWatchingTapTest` (5), `NeedsServerTest` (5).

Not covered by automated tests: the actual dimming and snackbar are Compose UI, and this repo has no Compose UI test suite yet — those were reasoned through rather than exercised.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TDhoJNqhHYVpmMDmUbUkwX)_


app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerViewModel.kt
```diff
@@ -157,24 +169,46 @@ class VideoPlayerViewModel(
...
             return
         }
         _nowPlaying.value = item
         candidates = resolveSources(id)
         candidateIndex = 0
         reachedReady = false
-        if (candidates.isEmpty()) {
-            _buffering.value = false
-            _error.value = "No playable source for this title"
+        val blocker = playbackBlocker()
```
jaxkodex, line 182: A blocked hop here corrupts the *next* episode's resume point.

`_nowPlaying` and `reachedReady` are both mutated (lines 178–181) **before** `playbackBlocker()` can refuse. On an autoplay hop into an undownloaded next episode with the server gone:

- the player is never told to stop, so E1 keeps playing;
- `_nowPlaying` already points at E2, and `reportProgressLoop` writes `player.currentPosition` / `player.duration` against **E2's id** every 10 s — E1's playhead saved as E2's resume position, locally and pushed to the server;
- `reachedReady = false` on line 181 also means E1's own STOPPED record is discarded.

The pre-existing `candidates.isEmpty()` branch had the same shape, but it was near-unreachable in practice; this branch is routine (it's the case the PR exists for), so the latent bug becomes a real one.

Suggest computing the blocker before publishing any state — resolve sources, run `playbackBlocker()`, and only then assign `_nowPlaying` / `candidateIndex` / `reachedReady` — or explicitly `player.stop()` and restore the previous `_nowPlaying` on the refusal path.

---

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerViewModel.kt
```diff
@@ -111,6 +113,16 @@ class VideoPlayerViewModel(
             } else {
                 _buffering.value = false
                 _error.value = error.message ?: "Playback failed"
+                // Every candidate exhausted on a server-backed stream usually means the server went
+                // away mid-playback (VPN dropped, walked off the LAN). Confirm in the background and
+                // swap in the message that names the real problem; the raw one stands until then.
+                if (needsServer(candidates)) {
```
jaxkodex, line 119: This background write to `_error` has no generation guard, so it can land on a different item.

`check()` can take up to `PING_TIMEOUT_MS` (3 s). If the user taps Up Next in that window and the next episode is a downloaded `file://` copy, `startItem` clears `_error` and playback starts fine — then this coroutine resolves and paints "Can't reach your server" over playing video. Nothing clears it afterwards, since the successful path already ran its `_error.value = null`.

Cheap fix: capture the item/attempt id before launching and drop the write if it no longer matches, or cancel the previous check job at the top of `startItem`.

---

app/src/main/java/pe/net/libre/mixtapehaven/data/network/ServerAvailability.kt
```diff
@@ -0,0 +1,125 @@
...
+    private fun markUnreachable() {
+        _reachable.value = false
+    }
+
+    private companion object {
+        /** How long a successful exchange stands in for a ping. */
+        const val FRESH_MS = 30_000L
+
+        /** Ping budget. Long enough for a slow LAN, short enough that a tap still feels answered. */
+        const val PING_TIMEOUT_MS = 3_000L
```
jaxkodex, line 123: A timed-out ping is treated as a definite "no", which turns this budget into a hard gate on playback.

`pingNow` maps `withTimeoutOrNull(...) ?: false` straight into `markUnreachable()`, and `playbackBlocker` refuses on that. So a *working* remote server that just answers slowly — cold DNS + TCP + TLS on a weak mobile connection, or a sleeping VPS — gets the user a "check your VPN" message for a title that would have streamed. The old behaviour at least tried and let ExoPlayer's own timeouts decide.

Two things compound it:

- Nothing calls `report(true)` on successful playback, so the 30 s fresh window is almost always stale by the time the player opens — meaning practically every non-downloaded play goes through this gate.
- 3 s is measured from lock acquisition, so a queued caller can wait considerably longer than the "tap still feels answered" budget the comment describes.

Worth considering: treat a timeout as inconclusive (leave `_reachable` alone and let playback proceed) versus a connection *refusal*, which is conclusive. And feeding `STATE_READY` back in via `report(true)` would keep the fresh window warm during a session for free.

---

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/home/HomeViewModel.kt
```diff
@@ -180,6 +207,33 @@ class HomeViewModel(
...
+     * it is confirmed with [ServerAvailability.check] before refusing: a server that came back
+     * (VPN reconnected) must play on the first tap, not after a manual reload.
+     */
+    fun resumeVideo(video: VideoItem, onResume: (String) -> Unit) {
+        val state = _state.value
+        if (canPlayNow(video.id, state.serverReachable, state.downloadedVideoIds)) {
+            onResume(video.id)
+            return
+        }
+        viewModelScope.launch {
```
jaxkodex, line 227: Each tap launches its own unguarded coroutine, with no in-flight state to show for it.

A dimmed card gives no feedback for up to 3 s, which is exactly the "did that do anything?" gap this PR is fixing — so the user taps again. Four taps means four coroutines: they serialize on `pingLock`, each re-pings on failure (`hasFreshSuccess()` is false after a failure, so the double-check doesn't collapse them), and each sends its own snackbar. The result is the same message queued four times over ~12 s.

The success case is worse in a different way: a late `onResume(video.id)` isn't checked against the current screen, so a ping that resolves after the user navigated elsewhere pushes the player on top of wherever they went.

Suggest a single in-flight job (`pendingResume?.cancel()` or a guard flag that also drives a spinner/disabled state on the card), and dropping the `onResume` call if the tap is no longer current.

---

jaxkodex (commented):
Reviewed the reachability gating across `ServerAvailability`, Home, and the video player.

The core design holds up well: separating "device has a network" from "the server answers" is the right call for a self-hosted Jellyfin, `needsServer` correctly keys off the `file://` uri that `VideoDownloadManager.localUriFor` produces, and folding real request outcomes into `report()` avoids a ping on the common path. The ping mutex and the second `hasFreshSuccess()` look under the lock are both correct, and `withTimeoutOrNull` + the inner `runCatching` don't swallow parent cancellation into a spurious `markUnreachable`.

Four findings inline, roughly in severity order:

1. **`VideoPlayerViewModel:182`** — a blocked autoplay hop leaves the previous item playing while `_nowPlaying` already points at the next one, so progress writes land against the wrong item id. This is the one I'd fix before merge.
2. **`VideoPlayerViewModel:119`** — the background `check()` writes `_error` with no generation guard, so a stale "can't reach your server" can be painted over a downloaded episode that's playing fine.
3. **`ServerAvailability:123`** — a timed-out ping is treated as conclusive, so a slow-but-working server blocks playback that would previously have succeeded.
4. **`HomeViewModel:227`** — one coroutine per tap, no pending UI: repeat taps queue duplicate snackbars, and a late `onResume` can push the player onto whatever screen the user moved to.

Also verified and clean: `ACCESS_NETWORK_STATE` is declared, `videoStreamCandidates` doesn't throw offline (so `playbackBlocker` is actually reachable), there are no other call sites of the two changed ViewModel constructors, and the new tests compile against the declared `kotlinx-coroutines-test` dependency.

---

PR #164, opened 2026-08-09
fix: stop battery drain from held wake locks and ungated polling

jaxkodex:
Started from a report of the phone heating up and the battery going down faster than usual. The first commit guessed at the video player; a battery-stats screenshot then showed **Mixtape at 11%, with 3 min of screen time against 1 hr 42 min in the background, playing downloaded music** — which redirected this to the audio path. Both are fixed here, but the WifiLock and position-poll fixes are the ones that match the reported symptom.

## What was draining

**A WifiLock held for local playback** (`PlaybackService`). The ExoPlayer was built with `setWakeMode(C.WAKE_MODE_NETWORK)` unconditionally. That mode adds a WifiLock on top of the CPU lock, holding the Wi-Fi radio out of power-save for as long as anything plays — right for a stream, pure waste for a file already on disk. Offline background playback is the longest, most common case where it buys nothing. The builder now takes `WAKE_MODE_LOCAL`, and a listener raises it per item only for an `http(s)` URI.

**The 500 ms position poll running with no reader** (`PlayerController`, `NowPlayingScreen`). #145 gated it on `isPlaying`, but `isPlaying` is exactly true during background playback, and the two flows it feeds exist only to drive the Now Playing progress bar. Two gaps: no gate on whether anything was collecting them, and `collectAsState` keeps collecting when the app is backgrounded, because stopping the Activity does not leave the composition. Now gated on `subscriptionCount` and collected with `collectAsStateWithLifecycle`, so it runs only while Now Playing is visible, foregrounded, and playing.

**The display pinned on by the video player** (`VideoPlayerScreen`). `keepScreenOn = true` was set in the `AndroidView` factory and never cleared, tying the screen lock to "the player screen is composed" rather than "something is playing" — so a paused video or a `Playback failed` message held the display lit indefinitely. Now driven from `playbackActive || buffering`.

**The video progress-report loop outliving its screen** (`VideoPlayerViewModel`). It woke every 10 s for the life of the ViewModel, which survives backgrounding, so it kept ticking after a pause, after a terminal error, and after a video ended with nothing queued. Now gated on `playbackActive`.

## Notes

- `playbackActive` is `playWhenReady && state in (READY, BUFFERING)`, not `Player.isPlaying`, so a rebuffer does not blank the screen mid-film. `|| buffering` additionally covers the pre-`prepare()` resolve window, where the player is `STATE_IDLE` — without it a slow resolve lets the display sleep, and the `ON_STOP` observer turns that into a pause the user never asked for.
- A stream stuck buffering still holds the screen lock. That is unchanged behaviour rather than something this PR fixes, and it is bounded in practice: a stalled load ends in an error within the source's read timeout, which releases it. A grace window was considered and rejected — it reintroduces a timer and needs a threshold nobody can pick well.
- Liveness detection for a killed session moved from the poll loop to `MediaController.Listener.onDisconnected`, since the loop is now subscriber-gated and would miss it in exactly the backgrounded case where the service is most likely to be reclaimed.
- `wakeModeForUri` is a pure seam alongside `resolvePlayables` / `shouldRefillQueue`, covered for the streamed, downloaded, and unknown-scheme cases. Unknown schemes resolve to `WAKE_MODE_LOCAL`: erring local costs a stream its WifiLock, erring network costs every offline listener battery for nothing.
- Merged `main` rather than rebasing. The one conflict was in `PlaybackService.onCreate`, where #162's session activity and this branch's wake-mode listener both landed — both kept. #163's `fail()` helper clears `_buffering` on every terminal load path, which is what the `|| buffering` screen lock relies on to release, so the two compose correctly.

## Testing

`./gradlew compileDebugKotlin testDebugUnitTest detekt` green — 120 unit tests, including three new `wakeModeForUri` cases.

**Not measured on a device.** Every fix here is reasoned from the code, and none of it has been checked against a real battery trace. The WifiLock is the largest of the four by some margin, but the honest next step is a charge cycle with the built APK before trusting any of these numbers.

### DIFF
--- a/app/src/main/java/pe/net/libre/mixtapehaven/data/playback/PlayerController.kt
+++ b/app/src/main/java/pe/net/libre/mixtapehaven/data/playback/PlayerController.kt
@@ -108,6 +108,30 @@ class PlayerController(
         }
     }
 
+    /**
+     * Session liveness, kept separate from the progress poll.
+     *
+     * A session that dies mid-playback sends no [Player.Listener] callback, so this used to be
+     * noticed only by the poll loop finding a null controller. That detection cannot live in the
+     * loop any more: the loop is gated on a Now Playing subscriber, which is absent in exactly the
+     * case the service is most likely to be reclaimed — backgrounded and under memory pressure.
+     * Left stuck, [isPlaying] would keep Home's mini-player showing Pause for a player that no
+     * longer exists, and tapping it would do nothing.
+     *
+     * A listener rather than a slow poll: this is an event the session already reports, and a
+     * second timer ticking through backgrounded playback is the cost this change set exists to
+     * remove.
+     */
+    private val controllerListener = object : MediaController.Listener {
+        override fun onDisconnected(controller: MediaController) {
+            // Ignore a stale controller already replaced by reconnect().
+            if (this@PlayerController.controller !== controller) return
+            controller.release()
+            this@PlayerController.controller = null
+            _isPlaying.value = false
+        }
+    }
+
     init {
         // Poll position/duration only while something is playing *and* a screen is displaying it.
         //
@@ -123,13 +147,9 @@ class PlayerController(
                 .collectLatest { shouldPoll ->
                     if (!shouldPoll) return@collectLatest
                     while (currentCoroutineContext().isActive) {
-                        val c = activeController()
-                        if (c == null) {
-                            // Controller lost mid-playback (service killed): stop polling and reset
-                            // the flag, else its conflated true would keep the loop from restarting.
-                            _isPlaying.value = false
-                            break
-                        }
+                        // Backstop only; [controllerListener] owns liveness. Still worth breaking
+                        // on, so a loop that outlives its controller stops rather than spinning.
+                        val c = activeController() ?: break
                         _durationMs.value = c.duration.coerceAtLeast(0)
                         _positionMs.value = c.currentPosition.coerceAtLeast(0)
                         delay(PROGRESS_INTERVAL_MS)
@@ -267,7 +287,9 @@ class PlayerController(
         if (connecting) return
         connecting = true
         val token = SessionToken(appContext, ComponentName(appContext, PlaybackService::class.java))
-        val future = MediaController.Builder(appContext, token).buildAsync()
+        val future = MediaController.Builder(appContext, token)
+            .setListener(controllerListener)
+            .buildAsync()
         future.addListener(
             {
                 connecting = false
--- a/app/src/main/java/pe/net/libre/mixtapehaven/data/playback/PlayerController.kt
+++ b/app/src/main/java/pe/net/libre/mixtapehaven/data/playback/PlayerController.kt
@@ -17,10 +17,12 @@ import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.currentCoroutineContext
 import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.collectLatest
+import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
 import pe.net.libre.mixtapehaven.data.diagnostics.DiagnosticsLog
@@ -107,26 +109,45 @@ class PlayerController(
     }
 
     init {
-        // Poll position/duration only while something is actually playing; while paused or idle
-        // the loop is suspended so the app does no periodic work (seeks are handled by the listener).
+        // Poll position/duration only while something is playing *and* a screen is displaying it.
+        //
+        // Gating on isPlaying alone still left the loop running through background playback: a
+        // locked phone playing a downloaded album has no Now Playing screen collecting these, so
+        // the main thread was woken twice a second for the length of the album to recompute state
+        // with no reader. Subscriber count is the honest signal — these two flows exist solely to
+        // drive a progress bar, so no subscriber means no reason to poll. Seeks are still reflected
+        // by the listener, and re-subscribing updates before its first delay, so a screen that
+        // comes back never shows a stale position.
         scope.launch {
-            _isPlaying.collectLatest { playing ->
-                if (!playing) return@collectLatest
-                while (currentCoroutineContext().isActive) {
-                    val c = activeController()
-                    if (c == null) {
-                        // Controller lost mid-playback (service killed): stop polling and reset
-                        // the flag, else its conflated true would keep the loop from restarting.
-                        _isPlaying.value = false
-                        break
+            combine(_isPlaying, positionSubscribers()) { playing, watched -> playing && watched }
+                .collectLatest { shouldPoll ->
+                    if (!shouldPoll) return@collectLatest
+                    while (currentCoroutineContext().isActive) {
+                        val c = activeController()
+                        if (c == null) {
+                            // Controller lost mid-playback (service killed): stop polling and reset
+                            // the flag, else its conflated true would keep the loop from restarting.
+                            _isPlaying.value = false
+                            break
+                        }
+                        _durationMs.value = c.duration.coerceAtLeast(0)
+                        _positionMs.value = c.currentPosition.coerceAtLeast(0)
+                        delay(PROGRESS_INTERVAL_MS)
                     }
-                    _durationMs.value = c.duration.coerceAtLeast(0)
-                    _positionMs.value = c.currentPosition.coerceAtLeast(0)
-                    delay(PROGRESS_INTERVAL_MS)
                 }
-            }
         }
     }
+
+    /**
+     * True while anything is collecting [positionMs] or [durationMs].
+     *
+     * Both are checked because a caller may want the duration alone; either one is a reader that
+     * would otherwise see a frozen value.
+     */
+    private fun positionSubscribers(): Flow<Boolean> =
+        combine(_positionMs.subscriptionCount, _durationMs.subscriptionCount) { position, duration ->
+            position + duration > 0
+        }
 
     /** Set the origin label for the next queue; call before [play] when starting a new source. */
     fun setSource(source: PlaybackSource) {
--- a/app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerScreen.kt
+++ b/app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerScreen.kt
@@ -134,10 +134,13 @@ fun VideoPlayerScreen(
 @Composable
 private fun PauseWhenScreenStops(onScreenStopped: () -> Unit) {
     val lifecycleOwner = LocalLifecycleOwner.current
-    val currentCallback by rememberUpdatedState(onScreenStopped)
+    // Read through .value rather than a `by` delegate. Both are the same read; the delegate is
+    // what static analysis flags as an unused variable, because the only use is a getValue call
+    // inside the observer lambda below and that is a hop it does not follow.
+    val currentCallback = rememberUpdatedState(onScreenStopped)
     DisposableEffect(lifecycleOwner) {
         val observer = LifecycleEventObserver { _, event ->
-            if (event == Lifecycle.Event.ON_STOP) currentCallback()
+            if (event == Lifecycle.Event.ON_STOP) currentCallback.value()
         }
         lifecycleOwner.lifecycle.addObserver(observer)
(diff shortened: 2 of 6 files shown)


app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerViewModel.kt
```diff
@@ -235,22 +259,28 @@ class VideoPlayerViewModel(
...
-                ready && !reportedPause -> {
-                    reportedPause = true
-                    reportProgress(paused = true)
-                }
+        _playbackActive.collectLatest { active ->
+            if (!active) {
+                // One report as playback stops, so the resume point lands. STATE_READY excludes the
+                // end-of-item and error transitions, which write their own STOPPED or show a
+                // message, and the initial idle emission before the first item is prepared.
+                if (player.playbackState == Player.STATE_READY) reportProgress(paused = true)
```
jaxkodex: **Pausing during a rebuffer now loses the paused report permanently.**

This branch is the only place the paused report can fire, and it only fires on the *edge* where `playbackActive` flips to false. `STATE_READY` is not guaranteed to hold at that edge:

1. Playing, `STATE_READY`, `playWhenReady = true`, flag is `true`.
2. Network hiccup → `STATE_BUFFERING`. `refreshPlaybackActive()` still yields `true` (that's the point of including BUFFERING), so no emission.
3. User taps pause → `onPlayWhenReadyChanged(false)` → flag emits `false`. `collectLatest` cancels the inner loop and enters this branch — but `player.playbackState` is still `STATE_BUFFERING`, so the guard skips the report.
4. ExoPlayer keeps loading while paused and settles to `STATE_READY` → `onPlaybackStateChanged` → `refreshPlaybackActive()` recomputes `false` again. `MutableStateFlow` conflates equal values, so **no emission**, so the collector is never re-entered.

Net: no paused report for that pause, and nothing will retry it. The old loop handled this because it ticked unconditionally — once the state settled to `READY` the next tick hit `ready && !reportedPause` and the report landed.

`onCleared()` still writes `STOPPED` when the user navigates away, so the common path is covered — but pause-then-background-then-process-death is exactly the case where `onCleared` never runs, and that's the case this report exists for. The resume point regresses to the last 10 s `PROGRESS` tick before the stall.

Suggest carrying the intent across the state transition rather than sampling it once, e.g. set a `pendingPauseReport` flag here when the state isn't `READY` yet, and flush it from `onPlaybackStateChanged(STATE_READY)` while `playWhenReady` is false. That keeps the `STATE_ENDED`/error exclusions the comment relies on.

---

app/src/main/java/pe/net/libre/mixtapehaven/data/playback/PlayerController.kt
```diff
@@ -107,27 +109,46 @@ class PlayerController(
...
+            combine(_isPlaying, positionSubscribers()) { playing, watched -> playing && watched }
+                .collectLatest { shouldPoll ->
+                    if (!shouldPoll) return@collectLatest
+                    while (currentCoroutineContext().isActive) {
+                        val c = activeController()
+                        if (c == null) {
+                            // Controller lost mid-playback (service killed): stop polling and reset
+                            // the flag, else its conflated true would keep the loop from restarting.
+                            _isPlaying.value = false
+                            break
```
jaxkodex: **This dead-controller recovery is now gated on a Now Playing subscriber, and it's the backgrounded case that loses it.**

The loop is no longer just a progress poller — as the original comment says, it's also the only detector for "service killed mid-playback", since `onIsPlayingChanged` never arrives for a session that dies. Adding `positionSubscribers()` to the gate means the loop is suspended precisely when there's no Now Playing screen up, which is when the service is most likely to be reclaimed (app backgrounded, low memory).

So: service dies while backgrounded → this `break` never runs → `_isPlaying` stays stuck `true`. `HomeViewModel.isPlaying` (`HomeViewModel.kt:62`) aliases the same flow and feeds the mini-player's icon (`HomeScreen.kt:134` → `NowPlayingControls`), so the user comes back to Home and sees a Pause button for a player that no longer exists. It self-heals only once they open Now Playing and the loop restarts.

Worth splitting the two concerns: keep the liveness check gated on `_isPlaying` alone, and gate only the `_positionMs`/`_durationMs` writes on subscriber count. A cheap version is a second `collectLatest(_isPlaying)` that polls `activeController()` on a slow interval purely for the null check, leaving the 500 ms loop subscriber-gated as written here.

---

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerScreen.kt
```diff
@@ -83,9 +84,12 @@ fun VideoPlayerScreen(
                     player = viewModel.player
                     setShowNextButton(false)
                     setShowPreviousButton(false)
-                    keepScreenOn = true
                 }
             },
+            // Only while something is actually playing. Pinning this on for as long as the screen
+            // is composed keeps the display — the phone's largest power draw — lit indefinitely
+            // behind a paused video, a "Playback failed" message, or a stream that never resolves.
+            update = { it.keepScreenOn = playbackActive },
```
jaxkodex: Two things about the cases this comment (and the PR description) claims.

**"a stream that never resolves" isn't actually covered.** `refreshPlaybackActive()` includes `STATE_BUFFERING`, and a stream stalled mid-load sits in `STATE_BUFFERING` with `playWhenReady` still `true` — so `playbackActive` is `true` and the display stays lit indefinitely, exactly as before. That's a deliberate design choice (the `_playbackActive` KDoc explains why a rebuffer must not blank the screen), and I don't think it's the wrong call — but it means one of the three cases listed here and in the PR body is still open. Worth either dropping the claim or bounding it, e.g. only treat `STATE_BUFFERING` as live for some grace window past the last `STATE_READY`.

**The initial-load window is now unprotected, and that interacts badly with `onScreenStopped`.** Before `prepare()` the player is `STATE_IDLE`, so `playbackActive` is `false` for the whole of `progressStore.resolvePlayback(id)` + `resolveSources(id)` — the second of which negotiates the transcode media-source id with the server. On a slow or flaky Jellyfin instance that's easily tens of seconds with nothing but a spinner on screen and no touch input to defer the timeout. If the display sleeps there, the `ON_STOP` observer above fires `viewModel.onScreenStopped()` → `player.pause()`, so the user unlocks to a paused player they never paused. Previously `keepScreenOn = true` in the factory covered this window.

`playbackActive || buffering` would close it, at the cost of re-opening the stalled-stream case — which is really the same grace-window question as above.

---

jaxkodex (commented):
Reviewed the three commits. The overall direction is right and the diagnosis in each commit message holds up — the screen wake lock really was pinned for the life of the composition, the video progress loop really did outlive the visible screen, and the 500 ms music poll really did run through backgrounded playback with no reader.

Verified locally on `ff3badb`: `./gradlew compileDebugKotlin testDebugUnitTest` is green, including the three new `wakeModeForUri` cases.

Three findings, all inline. The first is a behavioural regression; the other two are gaps between what the change claims and what it delivers.

1. **`VideoPlayerViewModel` — pausing during a rebuffer permanently loses the paused progress report.** The report only fires on the `playbackActive` false-edge, and `STATE_READY` doesn't necessarily hold at that edge. When the state later settles to `READY`, `refreshPlaybackActive()` recomputes `false`, `MutableStateFlow` conflates it, and the collector is never re-entered. The old unconditional loop caught this on its next tick.
2. **`PlayerController` — the dead-controller recovery is now gated behind a Now Playing subscriber.** That loop was doing double duty: progress polling *and* detecting a killed service (its own comment says so). Gating it on subscriber count suspends it exactly when the service is most likely to be reclaimed, leaving `_isPlaying` stuck `true` and the Home mini-player showing a Pause button for a dead player.
3. **`VideoPlayerScreen` — "a stream that never resolves" isn't covered**, because `STATE_BUFFERING` counts as active; and the pre-`prepare()` resolve window is now unprotected, where a display timeout trips `ON_STOP` → `onScreenStopped()` → `pause()`.

Checked and found clean:

- `wakeModeForUri` matches what the resolvers actually produce — `AppContainer` wires `resolveLocalFirst(track, manager::localUriFor, repository::audioStreamUrl)`, and `DownloadManager.localUriFor` returns `Uri.fromFile(...)`, so downloads are `file://` and streams are `http(s)://`. Nothing lands in a third bucket, and erring toward `WAKE_MODE_LOCAL` on an unknown scheme is the right default.
- `setWakeMode` from inside `onMediaItemTransition` is on the player's application thread, so the `verifyApplicationThread()` check is satisfied, and it doesn't re-enter listener dispatch.
- `positionSubscribers()` reads `subscriptionCount` on the backing `MutableStateFlow`s, and `asStateFlow()` shares that count with the read-only wrapper, so the exposed flows are counted correctly. No self-subscription loop, since it collects the counts rather than the values.
- `collectAsStateWithLifecycle` resolves transitively (no new dependency needed) and `STARTED` is the right floor — it's what actually drops the subscription on background, which is the whole point of commit 3.
- Hoisting the position tick into `PlaybackProgress` genuinely narrows the recomposition scope; nothing else in `NowPlayingScreen` reads position or duration.
- `onPlayWhenReadyChanged` is the right second callback — pausing mid-buffer moves only `playWhenReady`, so `onPlaybackStateChanged` alone would miss it.

---

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerViewModel.kt
jaxkodex: Confirmed, including the conflation at step 4 — that's what makes it permanent rather than merely late. Fixed in 59ecc9a, but not with the pending flag.

The guard was sampling a narrower state than the flag it's paired with. `refreshPlaybackActive()` treats `READY || BUFFERING` as live; the report guard treated only `READY` as reportable, so the mid-rebuffer edge fell through a gap between two conditions that should always have been the same one. Extracted that pair as `holdsLiveItem()` and used it in both places, so the edge cannot land outside it by construction.

That keeps the exclusions the comment relies on: `STATE_IDLE` covers both never-prepared and errored, `STATE_ENDED` covers end-of-item, and both still skip. A `pendingPauseReport` would work too, but it adds a second piece of state to keep in sync with the first, to paper over the two conditions having drifted apart — worth avoiding when making them one condition is the whole fix.

---

app/src/main/java/pe/net/libre/mixtapehaven/data/playback/PlayerController.kt
jaxkodex: Agreed on the diagnosis — the loop was doing two jobs and I gated both. Traced the consequence you describe and it holds: `playPause()` also no-ops on a null `activeController()`, so the stale Pause button isn't just cosmetic, tapping it does nothing.

Split in 59ecc9a, but event-driven rather than the slow poll. A dying session already reports itself through `MediaController.Listener.onDisconnected`, so liveness now hangs off that, wired via `MediaController.Builder.setListener`. It clears the controller and resets `_isPlaying` immediately instead of within one poll interval, and it costs nothing while idle.

I avoided the second `collectLatest(_isPlaying)` because it would be a timer ticking through backgrounded playback, which is the exact cost this branch exists to remove — reintroducing a slower one to recover a signal the platform already pushes seemed like the wrong trade.

The `activeController()` null check stays in the loop as a backstop, now a `?: break` since resetting `_isPlaying` there is the listener's job. Worth flagging one gap I did not close: `connect()` calls `future.get()` with no `try`/`catch`, so a failed connection throws on the main executor. Pre-existing and separate from this branch, so I left it — happy to take it in its own PR if you want it.

---

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerScreen.kt
jaxkodex: Both correct. Took `playbackActive || buffering` in 59ecc9a, and dropped the claim rather than bounding it.

On the resolve window: this is the worse of the two, because it doesn't just waste power, it produces a wrong state — the user unlocks to a player they never paused, and `_buffering` is already true across exactly that span (`startItem` raises it before `resolvePlayback`), so it's the signal that was already there.

On the grace window, I considered it and decided against. It reintroduces a timer, which is what this branch has been removing, and it would have to be tuned against a number nobody can pick well — too short blanks the screen during a slow-but-fine load, too long and it isn't doing much. It also isn't load-bearing: a genuinely stalled load ends in a load error within the source's read timeout, which drops the player to `STATE_IDLE` and clears `_buffering`, releasing the lock. So the unbounded case is narrower than "indefinitely" — it's "until the stream errors out". If that turns out to be wrong in practice on a flaky server, a grace window is the right follow-up, but I'd rather see it happen than pre-empt it.

The comment now says what it actually does, and I've rewritten the PR description too — it was still only describing the first commit, so the two fixes that address the reported symptom weren't mentioned anywhere in it.

---

PR #165, opened 2026-08-09
Fix the reachability gate's misattributions and dead taps

jaxkodex:
Follow-up to #163, acting on [its post-merge review](https://github.com/jaxkodex/mixtapehaven-android/pull/163#pullrequestreview-4891099337). Three of the four findings were real bugs; the fourth is answered in part, and the part I disagreed with is explained below. A [second review round](https://github.com/jaxkodex/mixtapehaven-android/pull/165#pullrequestreview-4891299289) on this branch found four more, all fixed in the second commit — summarised at the end.

## A blocked hop no longer misattributes the previous episode's playhead

[Finding 1](https://github.com/jaxkodex/mixtapehaven-android/pull/163#discussion_r3743421703). `prepareItem` published `_nowPlaying` (and reset `candidates`/`reachedReady`) *before* `playbackBlocker()` could refuse. On a user-driven hop — tap Up Next with the server gone — E1 kept playing while `_nowPlaying` already pointed at E2, and `reportProgressLoop`, which pairs the player's live playhead with whatever `_nowPlaying` holds, wrote E1's position as **E2's** resume point every 10s, locally and to the server.

Sources are now resolved and the blocker computed before anything is published, and a load that ends in an error calls `stopPlayback()` — so no stream keeps running behind the message and no playhead is left for the progress reports to misattribute. The same stop now covers the "item won't resolve" path, which had the same shape.

## The background reachability check can't paint over a later item

[Finding 2](https://github.com/jaxkodex/mixtapehaven-android/pull/163#discussion_r3743422255). The `check()` launched from `onPlayerError` wrote `_error` with no guard, so a check begun on a failed stream could resolve seconds later and stamp "Can't reach your server" over a downloaded episode that had since started playing fine — with nothing left to clear it. A `loadGeneration`, bumped by every `startItem`, is captured at launch and re-checked before the write.

## Repeat taps on a dimmed card cost one check, not one each

[Finding 4](https://github.com/jaxkodex/mixtapehaven-android/pull/163#discussion_r3743423052). At most one confirmation runs at a time: a repeat tap on the title already being checked is dropped, moving to a different title abandons the first, and `UiState.resumePendingId` drives a spinner and a "Checking your server…" meta line on the card — so the tap that used to look dead for the duration now shows its work. The navigation that follows a *deferred* tap is gated on Home's own back-stack lifecycle, so a late answer no longer pushes the player on top of whatever screen the user moved to.

## Fewer taps reach the gate, and the ones that do wait less

[Finding 3](https://github.com/jaxkodex/mixtapehaven-android/pull/163#discussion_r3743422729), partly. Three changes, all aimed at the false-refusal risk the review describes:

- A server-backed stream reaching its first `STATE_READY` calls `report(true)`. Bytes arriving is proof the server answers, it costs nothing, and it keeps the fresh window warm across a session so each autoplay hop no longer waits on its own ping.
- Concurrent callers now join the ping already in flight rather than serializing behind a mutex. A refusal leaves no fresh success to short-circuit on, so four taps used to mean four round trips and four timeouts back to back, the last tap waiting for all of them.
- The ping budget goes 3s → 5s, which is the cold DNS + TCP + TLS case on a weak mobile connection.

**What I did not change:** treating a timed-out ping as inconclusive. The case this whole class exists for — a LAN address from off the LAN, a VPN that is down — has nobody to send a refusal, so it *only* ever surfaces as a timeout. Letting playback proceed on one would put the black frame back for the exact scenario the feature targets. The trade-off is real, so the constant now documents it, and the fixes above narrow the window in which it can bite.

## Second review round

Four more findings, all real, all fixed in `0293ea1`:

1. **Two loads could be in flight at once.** `playNext` reads `_upNext` synchronously and suspends on the STOPPED record before reaching `startItem`, so two quick taps both captured the same non-null next; an ENDED racing a tap did the same with different ids — the interleaved-publish shape the first commit set out to close. Hops now go through `startLoad()`, which cancels the load in flight and joins it before starting, and `prepareItem` calls `ensureActive()` before publishing so a doomed load can't touch the player or write a STARTED for an episode nobody will see.
2. **The lifecycle gate was too wide.** It wrapped the synchronous resume too, and `RESUMED` is stricter than "the user can tap this" — a back stack entry sits at `STARTED` in split-screen and mid-transition, where every Continue watching tap would have been a silent no-op. `resumeVideo` now takes `stillInFront`, consulted only before navigating on a late answer, and Home passes `STARTED`.
3. **The pending-tap guard keyed on video id rather than the job.** A → B → A within one check window cleared the spinner out from under a still-running check. Now keyed on job identity, with the job started `LAZY` so its body can't observe `pendingResume` before it is assigned.
4. **`report(true)` could be fed by a buffered seek.** `onPlaybackStateChanged` sees every transition, and a seek back into buffered content reaches `READY` without touching the network — refreshing the freshness window on minutes-old evidence, which is how a dead server gets waved through. It now fires only on the first `READY` of a load.

## Testing

`./gradlew testDebugUnitTest detekt lint assembleDebug` — all pass. Three new `ServerAvailabilityTest` cases cover the sharing: concurrent checks share one *failing* ping, a later check starts a fresh one, and a caller that gives up doesn't cancel the answer the others are waiting on. The player and Home changes are in ViewModels whose collaborators (ExoPlayer, `JellyfinRepository`, `PlayerController`) aren't fakeable as constructed, so they're covered by review and manual reasoning rather than by unit tests.

https://claude.ai/code/session_01GZD4g2mLAHrCvr1rsA2Cea

### DIFF
--- a/app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/home/HomeScreen.kt
+++ b/app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/home/HomeScreen.kt
@@ -45,6 +45,8 @@ import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.semantics.Role
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.dp
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.compose.LocalLifecycleOwner
 import pe.net.libre.mixtapehaven.R
 import pe.net.libre.mixtapehaven.di.appViewModel
 import androidx.compose.foundation.lazy.LazyRow
@@ -84,6 +86,9 @@ fun HomeScreen(
     val nowPlayingTrack by viewModel.nowPlaying.collectAsState()
     val isPlaying by viewModel.isPlaying.collectAsState()
     val hasDownloads = state.onDevice.isNotEmpty()
+    // Home's own back stack entry, so "still in front of the user" survives navigation, not just
+    // the app being backgrounded.
+    val lifecycle = LocalLifecycleOwner.current.lifecycle
     val snackbarHostState = remember { SnackbarHostState() }
     LaunchedEffect(viewModel) {
         viewModel.snackbarMessages.collect { message -> snackbarHostState.showSnackbar(message) }
@@ -121,7 +126,7 @@ fun HomeScreen(
                     onOpenDownloads = onOpenDownloads,
                     onOpenSettings = onOpenSettings,
                     videoNav = videoNav,
-                    onResumeVideo = { viewModel.resumeVideo(it, videoNav.onResumeVideo) },
+                    onResumeVideo = resumeVideoAction(viewModel, lifecycle, videoNav.onResumeVideo),
                     onPlayTrack = { viewModel.playOnDevice(it); onOpenNowPlaying() },
                     onPlayAlbum = { viewModel.playAlbum(it); onOpenNowPlaying() },
                     onRetry = viewModel::load,
@@ -199,6 +204,23 @@ private data class HomeSectionActions(
 )
 
 /**
+ * A Continue watching tap, with the resulting navigation gated on [lifecycle].
+ *
+ * A title that needs the server is confirmed against it before the player opens, which is a round
+ * trip, and Home stays composed behind whatever the user opens in the meantime. Navigating on a
+ * late answer would drop the player on top of the screen they moved to, so the tap lapses instead.
+ */
+private fun resumeVideoAction(
+    viewModel: HomeViewModel,
+    lifecycle: Lifecycle,
+    onResumeVideo: (String) -> Unit,
+): (VideoItem) -> Unit = { video ->
+    viewModel.resumeVideo(video) { id ->
+        if (lifecycle.currentState.isAtLeast(Lifecycle.State.RESUMED)) onResumeVideo(id)
+    }
+}
+
+/**
  * The library sections of Home, in design order: saved tracks, Continue watching, Movies & shows,
  * and the Recently added grid.
  */
@@ -220,6 +242,7 @@ private fun HomeSections(
             videos = state.continueWatching,
             downloadedIds = state.downloadedVideoIds,
             serverReachable = state.serverReachable,
+            pendingId = state.resumePendingId,
             onVideoClick = actions.onResumeVideo,
         )
     }
@@ -320,7 +343,7 @@ private fun OnDeviceSection(
  *
  * With the server out of reach ([serverReachable] false) the rail keeps listing everything, but
  * titles without a saved copy are marked unplayable rather than left looking like the downloaded
- * ones.
+ * ones. [pendingId] is the one whose tap is currently being confirmed against the server.
  *
  * The plain title (no "See all") matches the design: the rail is the complete list.
  */
@@ -329,6 +352,7 @@ private fun ContinueWatchingSection(
     videos: List<VideoItem>,
     downloadedIds: Set<String>,
     serverReachable: Boolean,
+    pendingId: String?,
     onVideoClick: (VideoItem) -> Unit,
 ) {
     SectionHeader(title = "Continue watching")
@@ -339,6 +363,7 @@ private fun ContinueWatchingSection(
                 video = video,
                 downloaded = downloaded,
                 unavailable = !serverReachable && !downloaded,
+                pending = video.id == pendingId,
                 onClick = { onVideoClick(video) },
             )
         }
--- a/app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/home/HomeScreen.kt
+++ b/app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/home/HomeScreen.kt
@@ -204,20 +204,28 @@ private data class HomeSectionActions(
 )
 
 /**
- * A Continue watching tap, with the resulting navigation gated on [lifecycle].
+ * A Continue watching tap, with the navigation that follows a *deferred* one gated on [lifecycle].
  *
  * A title that needs the server is confirmed against it before the player opens, which is a round
  * trip, and Home stays composed behind whatever the user opens in the meantime. Navigating on a
  * late answer would drop the player on top of the screen they moved to, so the tap lapses instead.
+ *
+ * `STARTED` rather than `RESUMED`, because the question is whether Home is still the screen in
+ * front of the user, not whether it holds focus: a back stack entry sits at `STARTED` in
+ * split-screen and mid-transition, and refusing there would make live taps do nothing — the very
+ * thing this rail is meant to stop. Navigation-compose drops a covered entry below `STARTED`, so
+ * the case worth refusing is still refused.
  */
 private fun resumeVideoAction(
     viewModel: HomeViewModel,
     lifecycle: Lifecycle,
     onResumeVideo: (String) -> Unit,
 ): (VideoItem) -> Unit = { video ->
-    viewModel.resumeVideo(video) { id ->
-        if (lifecycle.currentState.isAtLeast(Lifecycle.State.RESUMED)) onResumeVideo(id)
-    }
+    viewModel.resumeVideo(
+        video = video,
+        stillInFront = { lifecycle.currentState.isAtLeast(Lifecycle.State.STARTED) },
+        onResume = onResumeVideo,
+    )
 }
 
 /**
--- a/app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/home/HomeViewModel.kt
+++ b/app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/home/HomeViewModel.kt
@@ -4,8 +4,10 @@ import android.util.Log
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CoroutineStart
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
+import kotlinx.coroutines.currentCoroutineContext
 import kotlinx.coroutines.async
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.flow.Flow
@@ -231,8 +233,13 @@ class HomeViewModel(
      * That check is slow enough to be tapped through, so one runs at a time: a repeat tap on the
      * title already being checked is dropped rather than queued (four taps used to mean four pings
      * and the same message four times), and moving to a different title abandons the first.
+     *
+     * [stillInFront] is asked only on that slow path, and only about navigating: an answer can
+     * arrive after the user has moved on, and opening the player over the screen they went to is
+     * worse than letting the tap lapse. The immediate path never consults it — a downloaded title
+     * resumes on the tap, whatever the screen is doing.
      */
-    fun resumeVideo(video: VideoItem, onResume: (String) -> Unit) {
+    fun resumeVideo(video: VideoItem, stillInFront: () -> Boolean = { true }, onResume: (String) -> Unit) {
         val state = _state.value
         if (canPlayNow(video.id, state.serverReachable, state.downloadedVideoIds)) {
             onResume(video.id)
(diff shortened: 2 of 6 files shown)


app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerViewModel.kt
```diff
@@ -147,6 +162,7 @@ class VideoPlayerViewModel(
      * unguarded — and a spinner turning forever reads worse than the black frame it replaced.
      */
     private suspend fun startItem(id: String) {
+        loadGeneration++
```
jaxkodex, line 197: `loadGeneration` is the right tool, but it only guards one of the two places that can land on the wrong episode. `prepareItem` never consults it before publishing `_nowPlaying`, `candidates`, or calling `prepareCurrentCandidate` — and nothing serializes `startItem` itself.

Two `startItem` coroutines can be in flight at once, because both callers launch into `viewModelScope` with no mutual exclusion:

- `playNext()` reads `val next = _upNext.value ?: return` **synchronously**, then launches and suspends on `progressStore.record(...)` before reaching `startItem`. `_upNext` is only cleared inside `startItem`, so two quick taps on the next button both capture the same non-null `next` and both run a full load for it: two `resolveSources` round trips, two `serverAvailability.check()` calls, two STARTED reports, and a second `prepareCurrentCandidate(startMs = resolved.positionMs)` that re-seeks a stream already playing.
- `onPlaybackEnded()` (STATE_ENDED) racing a next tap at the very end of an episode is the same shape with *different* ids — whichever `resolveSources` returns last wins on `candidates` while the other may have already won on `_nowPlaying`, which is exactly the playhead-misattribution class finding 1 is fixing.

Cheapest fix that reuses what's already here: capture `val generation = loadGeneration` after the increment and gate every publish in `prepareItem` on `generation == loadGeneration` (and skip the `prepare()` when it isn't). Better still, keep the load in a `Job` and `cancelAndJoin()` it at the top of `startItem`, so there is only ever one.

---

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/home/HomeScreen.kt
```diff
@@ -198,6 +203,23 @@ private data class HomeSectionActions(
...
+ * trip, and Home stays composed behind whatever the user opens in the meantime. Navigating on a
+ * late answer would drop the player on top of the screen they moved to, so the tap lapses instead.
+ */
+private fun resumeVideoAction(
+    viewModel: HomeViewModel,
+    lifecycle: Lifecycle,
+    onResumeVideo: (String) -> Unit,
+): (VideoItem) -> Unit = { video ->
+    viewModel.resumeVideo(video) { id ->
+        if (lifecycle.currentState.isAtLeast(Lifecycle.State.RESUMED)) onResumeVideo(id)
```
jaxkodex: Two things about this gate, both pushing toward `STARTED` instead of `RESUMED`.

**It also wraps the synchronous path.** `resumeVideoAction` replaces the whole `onResumeVideo` callback, so it gates the `canPlayNow` early return in `resumeVideo` too — the downloaded/reachable tap that returns before any coroutine runs. That tap never needed a lifecycle gate; it can only be blocked by it.

**`RESUMED` is stricter than "the user can tap this".** Compose delivers touches while the entry is `STARTED`, and a nav back stack entry sits at `STARTED`-but-not-`RESUMED` in real, interactive situations: split-screen or multi-window where the other pane holds focus and the platform doesn't grant multi-resume, and the frames of an in-progress transition. In those states every Continue watching tap is a silent no-op — the exact dead tap #163 set out to eliminate, reintroduced for the case that used to work instantly.

`STARTED` costs nothing on the case you're actually defending against: navigation-compose drops a back-stack entry below `STARTED` once another destination covers it, and backgrounding the app takes it to `CREATED`. So a late answer arriving after the user opened the player, or after they left the app, still lapses.

Narrower still if you want both: gate only the deferred callback (pass the lifecycle check into `resumeVideo`'s coroutine branch) and leave the synchronous resume ungated.

---

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/home/HomeViewModel.kt
```diff
@@ -217,20 +227,33 @@ class HomeViewModel(
...
+                if (serverAvailability.check()) {
+                    onResume(video.id)
+                    return@launch
+                }
+                diagnostics.log(TAG, "Blocked resume of ${video.id}: server unreachable, no saved copy")
+                _snackbarMessages.send(unavailableMessage(video.title, hasNetwork = serverAvailability.hasNetwork()))
+            } finally {
+                // Only if this tap is still the pending one: a tap on another title has already
+                // published its own wait, and clearing that would mark its card as idle mid-check.
+                _state.update { if (it.resumePendingId == video.id) it.copy(resumePendingId = null) else it }
```
jaxkodex: The guard keys on the video id, but the thing it needs to identify is the *job*. A → B → A within one check window defeats it:

1. Tap A: `pendingResume` = job1, `resumePendingId` = A.
2. Tap B: `job1.cancel()`, `resumePendingId` = B, `pendingResume` = job2.
3. Tap A again: `resumePendingId` is B so the dedup at the top doesn't fire; `job2.cancel()`, `resumePendingId` = A, `pendingResume` = job3.
4. job1's `finally` — cancellation is delivered at its suspension point inside `check()`, so it can run any time after step 2 — sees `resumePendingId == A` and nulls it.

job3 keeps checking with no spinner and no "Checking your server…" on the card, and the tap looks dropped again. Narrow window, small blast radius (navigation and the snackbar still work), but it's the same failure this block exists to prevent.

`if (pendingResume == coroutineContext[Job]) pendingResume = null` alongside the state clear — or clearing on job identity rather than id — closes it.

---

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerViewModel.kt
```diff
@@ -84,13 +84,23 @@ class VideoPlayerViewModel(
...
         override fun onPlaybackStateChanged(playbackState: Int) {
             when (playbackState) {
                 Player.STATE_BUFFERING -> _buffering.value = true
                 Player.STATE_READY -> {
                     reachedReady = true
                     _buffering.value = false
+                    // Bytes arriving over the network is proof the server answers, and it is free:
+                    // folding it in keeps the fresh window warm for the whole session, so the next
+                    // episode's gate can skip its ping instead of making every hop wait on one.
+                    if (needsServer(candidates)) serverAvailability.report(success = true)
```
jaxkodex: The premise — "bytes arriving is proof the server answers" — holds for the first `STATE_READY` and for every rebuffer, but `onPlaybackStateChanged` fires on every *transition*, and not all of them involve the network. A backward seek into already-buffered content goes `BUFFERING → READY` served entirely from ExoPlayer's buffer, and this then stamps `lastSuccessAtMs = now()` on evidence that is minutes old.

Consequence is small but points the wrong way: a user scrubbing back through a show whose server died keeps `hasFreshSuccess()` true indefinitely, so the next `check()` — including the gate on the autoplay hop — short-circuits to reachable and hands back the black frame the gate exists to prevent. It's the mirror of the false refusal you're fixing, and it's the one the class was written to stop.

`onPlaybackStateChanged` doesn't distinguish the two, but `Player.Listener.onPositionDiscontinuity` with `DISCONTINUITY_REASON_SEEK` does: set a flag there and skip the report on the `READY` that follows it. Or fold `report(true)` in only on the first `READY` per `startItem`, which already covers the stated goal of keeping the window warm across autoplay hops.

Low confidence on the impact, high on the mechanism — worth a note in the comment at minimum, since as written it reads as though only network bytes can reach this line.

---

jaxkodex (commented):
Reviewed the full diff and built the branch: `./gradlew testDebugUnitTest detekt lint` is green here, so the testing claim checks out.

**The three fixes land.** Finding 1 is genuinely closed — resolving sources and computing the blocker before anything is published, plus `stopPlayback()` on both error exits, means `reportProgressLoop` can't pair a live playhead with a `_nowPlaying` that moved on: after `stopPlayback()` the player is `IDLE`, so the `playbackState == STATE_READY` guard in the loop refuses to report at all. I checked that clearing `reachedReady` there doesn't cost a legitimate STOPPED report on the way out — both hop callers (`onPlaybackEnded`, `playNext`) finalize the outgoing item before `startItem` runs, so `onCleared`'s early return has nothing left to write. Finding 2's generation guard is correct for the path it covers. The three new `ServerAvailability` tests are real: the abandon case would fail if the ping ran in the caller's scope, which is the thing worth pinning.

**On what you didn't change:** I agree, and I don't think it's a close call. A LAN address from off the LAN has nobody to send an RST — it can only ever present as a timeout, so treating timeouts as inconclusive would make the gate a no-op for its own motivating case. Documenting the trade-off on the constant is the right resolution.

**Four findings**, none blocking, ordered by how much they matter:

1. `prepareItem` publishes without checking `loadGeneration`, and nothing serializes `startItem` — two loads can be in flight at once, which is the same misattribution shape as finding 1.
2. The lifecycle gate uses `RESUMED` and wraps the synchronous resume too, so an interactive-but-unfocused Home makes every Continue watching tap dead — including downloaded titles that never needed a check.
3. The `finally` guard keys on video id rather than job identity; A → B → A clears the spinner out from under a live check.
4. `report(true)` on `STATE_READY` can be fed by a buffered seek rather than network bytes, refreshing the freshness window on stale evidence.

Details inline.

---

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerViewModel.kt
jaxkodex: Confirmed both races and took the second option. `startLoad(block)` now owns every hop: it cancels the load in flight, and the new job `join()`s it before running, so only one load exists at a time. The cancel happens eagerly at call time rather than inside the new job — otherwise a third request that superseded the second before it ever ran would leave the first alive.

`playNext`'s double-tap and the ENDED-vs-tap race both go through it, and the STOPPED record that precedes each hop is inside the job, so a superseded hop's record is abandoned too — harmless, since the tap that replaced it reads the same still-unpublished `_nowPlaying` and records the same item.

I kept the generation counter for the one thing it still guards (the background check in `onPlayerError`, which lives outside the load) and added `ensureActive()` in `prepareItem` before the publish block. Serialization alone would already prevent the interleaving, since the joined ordering means a stale publish lands strictly before the new load starts — but without it a doomed load still calls `prepareCurrentCandidate` and writes a STARTED for an episode nobody will see.

---

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/home/HomeScreen.kt
jaxkodex: Agreed on both counts, and took the narrower option. `resumeVideo` gained a `stillInFront: () -> Boolean` that is consulted only in the deferred branch, immediately before navigating — the `canPlayNow` early return is untouched, so a downloaded title resumes on the tap whatever the entry's state is. Home passes `STARTED`.

The split-screen point is the one that decided it: gating an interactive tap on `RESUMED` would have reintroduced the dead tap for the case that already worked instantly, which is a worse trade than the late-navigation it was defending against.

---

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/home/HomeViewModel.kt
jaxkodex: You're right — the id identifies the tap, not the work. Now keyed on the job: the `finally` clears `pendingResume` and `resumePendingId` only when `pendingResume == currentCoroutineContext()[Job]`, so in your A → B → A trace job1 sees job3 in the field and leaves it alone.

One wrinkle worth noting: `viewModelScope` dispatches on `Main.immediate`, so an eagerly-started body runs up to its first suspension *before* `pendingResume = job` executes — and if `check()` ever returned without suspending, the whole body would finish while the field still held the previous job, leaving `resumePendingId` set with nothing to clear it. The job is started `LAZY` and `start()`ed after the assignment, so its body can't observe the field before it's written.

---

app/src/main/java/pe/net/libre/mixtapehaven/ui/screens/video/VideoPlayerViewModel.kt
jaxkodex: The mechanism is right and it points the wrong way, which makes it worth more than a comment: a seek back into buffered content reaches `READY` without a byte off the network, and stamping `lastSuccessAtMs` there is how a dead server gets waved through the gate — the exact failure this class was written to stop, just inverted.

Took the second option: `report(true)` fires only on the first `READY` of a load (`val firstReady = !reachedReady`, checked before the flag is set). That covers the whole stated goal — keeping the window warm across autoplay hops — without needing to classify later transitions at all. It also gives up the genuine evidence in a mid-playback rebuffer, which I'm happy to lose: over-reporting reachability is the direction that costs the user a black frame, under-reporting only costs a ping. The comment now says which transitions can reach the line and why the later ones don't count.

---

